PR #7, opened 2018-09-25
First implementation of workflow business rules

lennym:
For now the business rules are the same as they were before - i.e. there's no business rules - everything is still instantly resolved.

BUT there is now an interface into which we can apply hook logic to divert different change-requests to different statuses and resolve at different places in the workflow.


lennym:
Waiting for https://github.com/UKHomeOffice/application-container-platform-bau/issues/4235 before this can deploy.

PR #10, opened 2018-10-04
Create a whitelist for particular change requests

lennym:
Allow some change requests to procede directly to resolved state, without going via inspectors


lennym:
Probably going to write some tests for this as it feels kind of important. Feel free to merge now or wait.

PR #16, opened 2018-10-18
Revert workflow disable commit

wheelsandcogs:


lennym:
You'll likely want this first: https://github.com/UKHomeOffice/taskflow/pull/8

lib/hooks/resolve/index.js
```diff
@@ -31,12 +32,11 @@ module.exports = settings => {
...
-    //   // insta-resolve
-    //   if (model.status === 'new') {
-    //     return pollChangelog(response.MessageId);
-    //   }
-    // });
+      .then(response => {
+        // insta-resolve
+        if (model.status === 'new') {
+          return pollChangelog(response.MessageId);
+        }
```
lennym: I don't think this conditional should be there. Shouldn't we always wait for the message to resolve?

joefitter: Yeah you're probably right. I was only doing it for things that went straight from `new` to `done`, but thinking about it we probably also want to wait for the reply if a case is resolved by some other means

lennym: _nods_

lib/hooks/resolve/index.js
```diff
@@ -32,11 +33,5 @@ module.exports = settings => {
   return model => {
     return queue(model.data)
       .then(response => pollChangelog(response.MessageId));
```
wheelsandcogs: @lennym @joefitter so leave it like this?

joefitter: this comment seems like it's in response to something, although I don't know what that is :)

wheelsandcogs: Click the "show outdated" link above to reopen the previous convo.

joefitter: ah yeah, leave it without the `if`

joefitter: soz I didn't know this comment was in reference to removing that

wheelsandcogs: no worries :D 

Am I good to merge then?

PR #18, opened 2018-10-22
insta resolve PIL create tasks

joefitter:


lennym:
It might be better to treat each case as a separate condition rather than continually adding to that one if statement.

joefitter:
yeah I did that, then undid it as repeated the setting of the module state. Save those chars

joefitter:
@lennym owzat?

lennym:
Niiiice. 

I'll eventually want to go to model-by-model conditions, but this is fine for now. Just stops that one condition getting out of hand before it starts.

PR #19, opened 2018-10-23
Workflow now decorates cases with extra metadata when possible (subject & establishment)

wheelsandcogs:
When a new case is created, workflow will add a `subject` (profile id that the case is related to / owned by) and if sensible, an `establishmentId`.


lib/api.js
```diff
@@ -12,6 +13,8 @@ module.exports = settings => {
 
   const flow = Taskflow({ db: settings.taskflowDB });
 
+  flow.hook('pre-create', hooks.validate(settings));
```
wheelsandcogs: Is `pre-create` the right event?

wheelsandcogs:
Open to suggestions for a better hook name than `meta`?

lib/hooks/meta/index.js
```diff
@@ -0,0 +1,42 @@
...
+  const { Roles } = db(settings.db);
+
+  return model => {
+    const type = get(model, 'data.model');
+    const data = get(model, 'data');
+
+    switch (type) {
+      case 'pil':
+        data.subject = get(data, 'data.profileId'); // pil holder
+        data.establishmentId = get(data, 'data.establishmentId');
```
lennym: I don't think this will be sufficient. Those parameters won't be on all cases, only the ones we have now with a `create` action.

I think we need to load the data from the database based on the id provided (if an id is provided).

lib/hooks/meta/index.js
```diff
@@ -0,0 +1,42 @@
...
+        break;
+
+      case 'place':
+        return Promise.resolve()
+          .then(() => Roles.query().findOne({ type: 'pelh', establishmentId: get(data, 'data.establishmentId') }))
+          .then(role => {
+            data.subject = role.profileId; // establishment licence holder
+            data.establishmentId = get(data, 'data.establishmentId');
+          })
+          .then(() => model.update(data));
```
lennym: As above, the `establishmentId` _might_ not be on the payload if it's an amendment.

joefitter:
LGTM, I'd like to defer to @lennym though as workflow is his baby

lib/hooks/meta/index.js
```diff
@@ -0,0 +1,42 @@
...
+        data.subject = get(data, 'data.id'); // profile user
+        break;
+
+      // case 'project':
+      //   data.subject = ''; // project licence holder
+      //   data.establishmentId = get(data, 'data.establishmentId');
+      //   break;
+
+      case 'trainingModule':
+        data.subject = get(data, 'data.profileId');
```
lennym: I don't think we'll _ever_ need metadata on a training module since they'll always insta-resolve.

*If* we decide we do want metadata then the same logic as above will apply.

lennym:
tl;dr - it's assuming params are always present on the payload. We should read associations from the db instead.

lennym:
Hook name is fine by me.

PR #20, opened 2018-10-23
resolve PIL updates if not active

joefitter:


lib/hooks/create/index.js
```diff
@@ -3,16 +3,26 @@ const { get } = require('lodash');
...
+
+      // inactive PIL updates
+      if (status !== 'active' && newStatus !== 'active') {
+        return model.setStatus('resolved');
+      }
+
+      // deactivating PIL
+      if (status === 'active' && newStatus !== 'active') {
+        return model.setStatus('resolved');
+      }
```
lennym: I think this is a bit of a long-winded way to say that the action we're trying to undertake is a revocation. I might suggest directly querying the `action` property of the new case data rather than inferring from previous and new state.

i.e. to revoke a PIL send the following payload:

```js
{
  model: 'pil',
  id: 'abc123',
  action: 'revoke'
}
```

And then determine if it's a revocation from that.

Similarly for granting a licence, use an action of `grant` rather than a particular flavour of `update`.

joefitter: 👍

PR #24, opened 2018-10-26
Decorate tasks with establishments, profiles, and next steps

wheelsandcogs:
When a list of tasks (cases) is fetched from the taskflow database, we need to provide some extra data to make them actually useful in the frontend.

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -24,6 +24,7 @@ module.exports = settings => {
 
   flow.hook('create', hooks.create(settings));
 
+  flow.hook('pre-status:*:autoresolved', hooks.resolve(settings));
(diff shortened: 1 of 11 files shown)


lib/api.js
```diff
@@ -20,6 +21,25 @@ module.exports = settings => {
...
+
+    if (res.body.data) {
+      let tasks = res.body.data;
+      console.log(tasks);
+
+      tasks = tasks.map(task => {
+        task.nextSteps = getNextSteps(task.data.model, task.status);
+      });
+
+      res.body.data = tasks;
```
wheelsandcogs: haven't tested this yet!

PR #25, opened 2018-10-30
auto-resolve exemptions and certifcates, removed training modules

joefitter:


lennym:
For mildly improved backwards compatibility I'd suggest in future adding the model types but not necessarily removing the old ones. This is probably fine now though.

PR #26, opened 2018-11-05
Don't fall over if there is no flow defined for model

wheelsandcogs:


lennym:
This is the first time I've seen the work around these flows.

I'd be keen to see a default flow defined, which then is extended by particular models where applicable.

That way we don't risk dumping particular requests into black holes.

PR #27, opened 2018-11-05
Use a constant for the autoresolved status

wheelsandcogs:


lennym:
I'm not sure I understand why this one status is a constant, and the others are still hard-coded.

wheelsandcogs:
`autoresolved` feels like a special status - 
 * it will be used across all model flows, which might not be true of other statuses (e.g. `ntco`)
 * it's the only one that's set in lots of different cases

I mainly wanted it to stand out so we don't set `resolved` when we actually mean `autoresolved`.

PR #29, opened 2018-11-06
Provide a default flow

wheelsandcogs:


lib/flow/default.js
```diff
@@ -0,0 +1,16 @@
...
+  autoresolved: [],
+
+  // applicant needs to make ammends and re-submit to licensing
+  applicant: ['licensing'],
+
+  // licensing can return to applicant for ammends, refer to inspector, or grant the licence
+  licensing: ['applicant', 'inspector', 'resolved'],
+
+  // inspector can recommend to licensing that they approve or reject the application
+  inspector: ['licensing'],
```
lennym: I don't think this is right if we're using the flow to describe the available future states of a change-request. For example, if a licensing officer has the option to send to one of three states, then the inspector should also have multiple options - else there is no decision to make.

Given this, it might be better to describe the statuses as the "state" of the change-request, rather than the "person" responsible for moving it out of that state - i.e. "returned to applicant", "awaiting triage", "awaiting recommendation", "recommended" etc.

wheelsandcogs: Ok lets discuss when I get to the office.

PR #32, opened 2018-11-08
Add seeds for cases

lvnhmd:
I just added some basic data to cases to be able to test datatable on user dashboard, @wheelsandcogs not sure if you'd agree how I structured the seed folder - it is based on asl-schema, but please do let me know your thoughts.

### DIFF
--- a/seeds/data/cases.json
+++ b/seeds/data/cases.json
@@ -1,546 +1,546 @@
 [
-    {
-        "status": "with-ntco",
-        "data": {
-            "id": "82b562d2-14d8-40fe-8a78-1ee313cc0a48",
-            "data": {
-                "profileId": "5b7bad13-f34b-4959-bd08-c6067ae2fcdd",
-                "establishmentId": 8201
-            },
-            "model": "pil",
-            "action": "grant",
-            "subject": "5b7bad13-f34b-4959-bd08-c6067ae2fcdd",
-            "changedBy": "5b7bad13-f34b-4959-bd08-c6067ae2fcdd",
-            "establishmentId": 8201
-        },
-        "updated_at": "2018-11-08 11:23:55.981023+00",
-        "created_at": "2018-11-08 11:23:55.981023+00"
+  {
+    "status": "with-ntco",
+    "data": {
+      "id": "82b562d2-14d8-40fe-8a78-1ee313cc0a48",
+      "data": {
+        "profileId": "5b7bad13-f34b-4959-bd08-c6067ae2fcdd",
+        "establishmentId": 8201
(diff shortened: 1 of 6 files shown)


lennym:
Are the referenced ids in the case seeds fixed against other _seeded_ ids?

That is, if we're referring to a thing by its id, then is that id hard-coded into the seed for that thing? If not, then it'll need to be or the references will break when the seeds run and ids are regenerated.

joefitter:
@lennym yes, the profileId is taken from asl-schema seeds, it's actually _your_ id.

seeds/data/cases.json
```diff
@@ -0,0 +1,546 @@
...
+            "model": "pil",
+            "action": "grant",
+            "subject": "5b7bad13-f34b-4959-bd08-c6067ae2fcdd",
+            "changedBy": "5b7bad13-f34b-4959-bd08-c6067ae2fcdd",
+            "establishmentId": 8201
+        },
+        "updated_at": "2018-10-26 11:23:55.981023+00",
+        "created_at": "2018-11-08 11:23:55.981023+00"
+    }
+]
```
joefitter: spacing is borked in this file

seeds/tables/cases.js
```diff
@@ -0,0 +1,15 @@
+const uuidv4 = require('uuid/v4');
+const cases = require('../data/cases.json');
+
+module.exports = {
+  populate: knex => {
+    return Promise.all(cases.map(c => {
+      return knex('cases')
+        .insert({
+          id: uuidv4(),
```
joefitter: why do we have this? Knex is set to auto generate ids.

joefitter: just do `.insert(cases)`

lvnhmd: We have uuid because I was getting error: null value in column "id" violates not-null constraint,
the db is taskflow - probably it's not configured

joefitter: we should fix the issue, not try and code a workaround

joefitter: ie, remove the notNull constraint

lvnhmd: hmm, is that a fix for primary key ?

joefitter: ¯\_(ツ)_/¯ not sure what you mean :)

lvnhmd: actually, I need to add a default - .defaultTo(knex.raw('uuid_generate_v4()')); - did not realise that

joefitter: yep!

PR #33, opened 2018-11-14
Make the case decorator work for single cases

wheelsandcogs:


joefitter:
My initial thought would be we should be using Promises, just for the sake of consistency. We use promises literally everywhere else in the app so it kinda bugs me that this is using an async function

lib/middleware/case-decorator.js
```diff
@@ -2,29 +2,40 @@ const { uniq } = require('lodash');
...
     taskCase.nextSteps = flow.getNextSteps(taskCase.status);
 
     return taskCase;
   });
 
+  if (req.case) {
+    req.case = cases[0];
+  } else {
+    req.cases = cases;
+  }
```
joefitter: this also bugs me a bit, it would be better if case decorator took a case and decorated it, let the singular/array stuff be the concern of taskflow where it is called

PR #34, opened 2018-11-20
* Enable snake mapper when running seeds

lvnhmd:
* Clean up


lennym:
What difference does this make?

Not arguing with it, just trying to understand what it does.

PR #39, opened 2018-11-21
Check status changes are valid for the flow and contain the correct payload

wheelsandcogs:
~Needs https://github.com/UKHomeOffice/taskflow/pull/19 merging first so we can bump the version here.~

Done.

### DIFF
--- a/lib/flow/index.js
+++ b/lib/flow/index.js
@@ -1,5 +1,5 @@
-const { get } = require('lodash');
-
+const { get, values } = require('lodash');
+const status = require('./status');
 const {
   resubmitted,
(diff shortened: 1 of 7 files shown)


lib/flow/index.js
```diff
@@ -35,7 +35,12 @@ const getNextSteps = (currentStepId) => {
   return get(flow, currentStepId, []);
 };
 
+const getAllSteps = () => {
+  return values(status);
```
lennym: FYI `Object.values` will do this without need for lodash. Not worth changing now, but worth being aware of.

PR #44, opened 2018-11-28
AS-805: Show the correct tasks in the task list depending on user role

wheelsandcogs:
Scopes the tasklist to the current user.

The cases a user sees depends upon their role at the establishments they are part of.


lib/router/tasks.js
```diff
@@ -0,0 +1,62 @@
+const Case = require('@ukhomeoffice/taskflow/lib/models/case');
+const { Router } = require('express');
+const router = Router({ mergeParams: true });
+
+const roleQueryBuilders = {
+  applicant: require('../query-builders/applicant'),
+  ntco: require('../query-builders/ntco')
+};
```
joefitter: Not sure I see the benefit of creating an object to contain 2 required files - the pattern we use elsewhere is to have an index.js in the root of the query-builder directory and destructure the required modules

joefitter: a la
```js
const { applicant, ntco } = require('../query-builders');
```

lib/query-builders/applicant.js
```diff
@@ -0,0 +1,24 @@
...
+
+module.exports = {
+  outstanding: (queryBuilder, profile) => {
+    queryBuilder.whereJsonSupersetOf('data', { subject: profile.id })
+      .andWhere('status', returnedToApplicant.id);
+  },
+
+  inProgress: (queryBuilder, profile) => {
+    queryBuilder.whereJsonSupersetOf('data', { subject: profile.id })
+      .whereNotIn('status', [autoResolved.id, resolved.id, rejected.id, withdrawn.id, returnedToApplicant.id]);
```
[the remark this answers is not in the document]
wheelsandcogs: The status strings in the db aren't quite identical to the name of the object (e.g. `autoResolved` vs `autoresolved` and `referredToInspector` vs `referred-to-inspector`), and the column is a string so the `whereNotIn()` needs the strings rather than the objects.

The reason for using the status objects everywhere is if there's a typo it won't actually run, rather than just silently being a bug as would happen with a hardcoded string.

We could write a function to return the id but then you'd have to call the function for each object so I'm not sure that's an improvement?

lib/router/tasks.js
wheelsandcogs: This is missing the other ones that exist - I can add an index to the queryBuilders dir instead?

joefitter: yeah it's more just a comment, not a blocker, we use an index.js file in most other places like this, IMO it would be nicer, but I also don't actually care that much

lib/query-builders/applicant.js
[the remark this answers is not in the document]
joefitter: yeah as you were then, as below just a comment but if there's a valid reason for it then that's fine by me 👍

lib/router/tasks.js
wheelsandcogs: better?

joefitter: 🤗

lib/router/tasks.js
```diff
@@ -0,0 +1,59 @@
...
+            userRoles.map(role => {
+              query.orWhere(builder => queryBuilders[role].completed(builder, profile));
+            });
+            break;
+        }
+
+        query = Case.orderBy({ query, sort: req.query.sort });
+        query = Case.paginate({ query, ...req.query });
+
+        // console.log(query.toString());
```
joefitter: if this is no longer a WIP could you remove this commented console.log please :)

lennym:
I don't think there's any known case where roles _other than_ NTCO will have assigned tasks, so I'd rather not have a load of mostly empty code files to cover cases that probably won't ever exist.

lib/api.js
```diff
@@ -33,8 +33,7 @@ module.exports = settings => {
 
   flow.migrate();
 
-  app.get('/', excludeAutoResolved);
-
+  app.use('/tasks', tasksRouter(flow));
```
wheelsandcogs: I added this route to provide a way to hit the standard taskflow `/` route if we still want it - do we still want it? If not I'll just make this `/`.

lennym: I don't think there's a good reason for us to want to go to the underlying '/' route, so I'd mount this on the root.

lib/query-builders/applicant.js
```diff
@@ -0,0 +1,24 @@
...
+  inProgress: (queryBuilder, profile) => {
+    queryBuilder.whereJsonSupersetOf('data', { subject: profile.id })
+      .whereNotIn('status', [autoResolved.id, resolved.id, rejected.id, withdrawnByApplicant.id, returnedToApplicant.id]);
+  },
+
+  completed: (queryBuilder, profile) => {
+    queryBuilder.whereJsonSupersetOf('data', { subject: profile.id })
+      .whereIn('status', [resolved.id, rejected.id, withdrawnByApplicant.id]);
+  }
+};
```
lennym: I _really_ like this. Nice work.

lib/router/tasks.js
```diff
@@ -0,0 +1,58 @@
...
+      })
+      .then(cases => {
+        const send = taskflow.responder(req, res);
+        return send(cases.results, { count: cases.total });
+      })
+      .catch(next);
+  });
+
+  return router;
+};
```
lennym: We'll probably want two versions of this file, with a branch pretty early on for external users and ASRU users.

I'm happy to do that now or kick it down the road as you prefer.

lib/util.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  getEstablishmentsWhereUserIsRole: (role, profile) => {
```
lennym: The Ronseal school of function naming.

PR #46, opened 2018-12-04
Tasklist integration tests

wheelsandcogs:
This adds support for running integration tests against workflow.

It uses the `taskflow-test` and `asl-test` (`workflow-asl-test` on `localhost`) databases.

### DIFF
--- a/scripts/migrate.js
+++ b/scripts/migrate.js
@@ -1,24 +1,9 @@
-/* eslint-disable implicit-dependencies/no-implicit */
-const knex = require('knex');
 const Taskflow = require('@ukhomeoffice/taskflow');
-const dbConfig = require('../knexfile');
+const config = require('../config');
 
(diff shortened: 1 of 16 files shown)


scripts/migrate.js
```diff
@@ -0,0 +1,13 @@
...
+const migrate = () => {
+  return Promise.resolve()
+    .then(() => Taskflow({ db: config.taskflowDB }).migrate())
+    .catch(err => {
+      console.log(err);
+      process.exit(1);
+    });
+};
+
+migrate();
```
lennym: If I were being pernickety I'd point out that there's no need to wrap code that's executed exactly once in a function. Good job I'm not.

lennym:
It's quite hard to see this and not come up with a lot of ideas of edge cases and situations that we should have test cases for, but as a start this is great.

PR #47, opened 2018-12-04
Load profile from database not API

lennym:
The API attepts to create a profile if it can't find one, which involves a call to workflow. If workflow then tries to load the profile from the API then this causes an infinite recursive loop of the two APIs calling one another. That is bad.

Instead load the user's profile from the database directly on workflow requests.

### DIFF
--- a/lib/middleware/profile.js
+++ b/lib/middleware/profile.js
@@ -1,0 +1,18 @@
+module.exports = settings => {
+  const { Profile } = settings.models;
+  return (req, res, next) => {
+
+    if (req.user.profile) {
+      return next();
+    }
+    return Profile.query()
+      .findOne({ userId: req.user.id })
+      .eager('[roles, establishments]')
+      .then(profile => {
+        req.user.profile = profile;
+      })
+      .then(() => next())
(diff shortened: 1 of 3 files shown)


lib/middleware/profile.js
```diff
@@ -0,0 +1,18 @@
+module.exports = settings => {
+  const { Profile } = settings.models;
+  return (req, res, next) => {
+
+    if (req.user.profile) {
+      return next();
+    }
```
lennym: This is in order to not totally screw @wheelsandcogs's integration tests which will set a profile manually.

lib/middleware/profile.js
```diff
@@ -0,0 +1,18 @@
...
+    return Profile.query()
+      .findOne({ userId: req.user.id })
+      .eager('[roles, establishments]')
+      .then(profile => {
+        req.user.profile = profile;
+      })
+      .then(() => next())
+      .catch(next);
+
```
joefitter: cool whitespace, bro

lennym: Thanks.

PR #50, opened 2019-01-17
Structure of comments has changed location slightly

wheelsandcogs:


lib/hooks/validate/payload.js
```diff
@@ -6,7 +6,7 @@ module.exports = () => {
   return model => {
     const next = getAllSteps().find(step => step.id === model.meta.next);
 
-    if (next.commentRequired && (!has(model.meta.payload, 'comment') || isEmpty(model.meta.payload.comment))) {
+    if (next.commentRequired && (!has(model.meta.payload.meta, 'comment') || isEmpty(model.meta.payload.meta.comment))) {
```
joefitter: this looks fragile - I'd go for a `_.get`

lennym: Good call.

PR #53, opened 2019-01-22
Add query builder for establishment admin users

lennym:
Establishment admin users should be able to see any tasks relating to their establishment.


lennym:
I've created a new ticket for unit test coverage for all of these query builders. Feel free to either wait, or merge this as you see fit.

PR #56, opened 2019-01-29
Change the task actions depending upon who is viewing the task

wheelsandcogs:
Waiting on https://github.com/UKHomeOffice/taskflow/pull/25

We should also use a similar method to validate case status changes.

### DIFF
--- a/lib/flow/get-next-steps-for-user.js
+++ b/lib/flow/get-next-steps-for-user.js
@@ -2,7 +2,13 @@ const flow = require('../flow');
 const { withdrawnByApplicant } = require('../flow/status');
 const queryBuilder = require('../query-builders');
 
+const userCreatedThisTask = (c, profile) => c.data.changedBy === profile.id;
+
 module.exports = (c, profile) => {
+  if (flow.autoForwards(c.status)) {
+    return flow.getNextSteps(c.status);
+  }
+
   // if the task is outstanding for the current user then they can action it
   return queryBuilder({ profile, progress: 'outstanding' })
     .andWhere('cases.id', c.id)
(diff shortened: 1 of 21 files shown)


lib/flow/get-next-steps-for-user.js
```diff
@@ -0,0 +1,23 @@
...
+        return flow.getNextSteps(c.status);
+      }
+
+      // if the task is in progress for the current user, then they can withdraw it
+      return queryBuilder({ profile, progress: 'inProgress' })
+        .andWhere('cases.id', c.id)
+        .then(cases => cases.results.length > 0)
+        .then(canWithdrawTask => {
+          return canWithdrawTask ? [withdrawnByApplicant] : [];
+        });
```
lennym: I don't think the logic here is entirely complete. Only non-ASRU users can withdraw tasks _unless_ they also created them.

lib/hooks/validate/status.js
```diff
@@ -1,14 +1,27 @@
...
 
-    if (!validNextSteps.includes(model.meta.next)) {
-      throw new BadRequestError('Invalid status change');
-    }
+    return Promise.resolve()
+      .then(() => {
+        // resubmitting an application triggers an auto-status change, don't need to validate against the user
+        if (currentStatus === 'resubmitted' && [withNtco.id, withLicensing.id].includes(nextStatus)) {
+          return;
+        }
```
lennym: I'd just add these to the available next steps for the status of `resubmitted` in the same way as every other status and allow the logic to be the same here.

This is doing the exact same thing as every other status, but in a _special_ way, for no reason that I can see.

lennym: Actually, now I possibly see - is this because once it _has_ a status of resubmitted then it no longer appears in the user's "outstanding" list, and so validating the ability of the user to modify its status fails?

lib/query-builders/inspector.js
```diff
@@ -16,6 +17,7 @@ module.exports = {
 
   inProgress: (queryBuilder, profile) => {
     queryBuilder.whereIn('status', [
+      ntcoEndorsed.id,
```
lennym: I don't _think_ we need this change, but nothing bad happens by virtue of having it.

PR #57, opened 2019-01-31
Changes to name or dob of a named person or licence holder need to be approved by licensing

wheelsandcogs:


lennym:
Maybe add a unit test or two and this is 👌

PR #61, opened 2019-02-01
Add endpoint for fetching tasks associated with a model

wheelsandcogs:


lib/router/model-tasks.js
```diff
@@ -0,0 +1,30 @@
...
+  router.get('/:modelId', (req, res, next) => {
+    Promise.resolve()
+      .then(() => {
+        return Case.query()
+          .whereJsonSupersetOf('data', { id: req.modelId })
+          .whereNotIn('status', closed());
+      })
+      .then(cases => {
+        const send = taskflow.responder(req, res);
+        return send(cases, { count: cases.length });
```
lennym: The count _should_ only ever be 0 or 1 so is _probably_ redundant. No harm done though.

lib/router/model-tasks.js
```diff
@@ -0,0 +1,30 @@
...
+
+module.exports = taskflow => {
+  const router = Router({ mergeParams: true });
+
+  Case.db(taskflow.db);
+
+  router.param('modelId', (req, res, next, modelId) => {
+    req.modelId = modelId;
+    next();
+  });
```
lennym: It doesn't really _matter_, but as a point of style this `router.param` is pretty much redundant as it's not actually doing anything with the parameter apart from copying it to another part of the request.

I'd generally preserve this kind of code for where there is some kind of parameter validation/processing that can be consolidated across multiple routes that use the parameter. Since there is only one route, and the code doesn't do anything then I'd get rid of it.

PR #62, opened 2019-02-04
if asru user updates restriction, update the case data

joefitter:


lib/hooks/update-data/place.js
```diff
@@ -0,0 +1,14 @@
+const { get, set } = require('lodash');
+
+module.exports = (settings, model) => {
+  const isAsru = get(model, 'meta.user.profile.asruUser');
+  const payload = get(model, 'meta.payload');
```
lennym: This will need to default to `{}` if we don't want it to break when pulling `restrictions` off it later.

lib/hooks/update-data/place.js
```diff
@@ -0,0 +1,14 @@
...
+
+  if (isAsru && payload.restrictions) {
+    const data = get(model, 'data');
+    set(data, 'data.restrictions', payload.restrictions);
+    return model.update(data);
+  }
```
lennym: I don't think this makes it possible to remove all the restrictions? We might need to address that.

Possibly not in this PR though.

PR #65, opened 2019-02-18
Add profile data to the activity log "changedBy" props

wheelsandcogs:
Decorate the "changedBy" profile id with the actual profile data so we don't have to request it later.

Waiting on https://github.com/UKHomeOffice/taskflow/pull/29 and a bump to taskflow

### DIFF
--- a/lib/decorators/metadata.js
+++ b/lib/decorators/metadata.js
@@ -10,19 +10,11 @@ module.exports = settings => {
     const promises = [
       c.data.establishmentId && cache.query(Establishment, c.data.establishmentId),
       c.data.subject && cache.query(Profile, c.data.subject),
-      c.data.changedBy && cache.query(Profile, c.data.changedBy),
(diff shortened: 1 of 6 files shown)


lib/decorators/metadata.js
```diff
@@ -5,16 +5,24 @@ module.exports = settings => {
...
     ];
 
     return Promise.all(promises)
-      .then(([establishment, subject, changedBy]) => {
+      .then(([establishment, subject, changedBy, logChangedBys]) => {
+        let activityLog = [];
+
+        activityLog = c.activityLog && c.activityLog.map(log => {
+          log.changedBy = logChangedBys.find(profile => profile && profile.id === log.changedBy) || {};
+          return log;
```
lennym: These will be in order, so there's no need for the additional `find`
```js
const activityLog = c.activityLog && c.activityLog.map((log, i) => {
  return { ...log, changedBy: logChangedBys[i] };
});
```

(Note: there's also no need for the `let activityLog`, it can be `const` at the point of assignment)

PR #72, opened 2019-02-27
added middleware to skip taskflow for project version updates

joefitter:


lib/api.js
```diff
@@ -8,6 +8,7 @@ const modelTasksRouter = require('./router/model-tasks');
 const decorators = require('./decorators');
 
 const profile = require('./middleware/profile');
+const resolveProjectVersionUpdates = require('./middleware/resolve-project-version-updates');
```
lennym: Java called, it wants its function names back.

lib/middleware/resolve-project-version-updates.js
```diff
@@ -0,0 +1,13 @@
+const { get, noop } = require('lodash');
+const resolve = require('../hooks/resolve');
+
+module.exports = settings => (req, res, next) => {
+  const model = get(req.body, 'model');
+  const action = get(req.body, 'action');
+  if (model === 'projectVersion' && action === 'update') {
+    return resolve(settings)({ data: req.body, update: noop })
```
lennym: We'll want to move this outside of the middleware invocation, so the hook is initialised with the settings exactly once.

```js
module.exports = settings => {
  const hook = resolve(settings);
  return (req, res, next) => {
    ...
  }
}
```

lib/api.js
joefitter: when you say "Java", do you mean @wheelsandcogs ?

lib/middleware/resolve-project-version-updates.js
joefitter: yes, good point

lib/api.js
lennym: :trollface:

wheelsandcogs: Nice function name 👍

PR #76, opened 2019-02-28
Fix bug with permissions query builder

lennym:
Refactors the query builder to group all of the `orWhere` calls for a users roles under a single where clause. The previous version added the `orWhere` to the top level query, which meant that the code in `./lib/flow/get-next-steps-for-user.js` that used this in conjunction with an `andWhere(id)` was not applying the `andWhere` to the entire query, but only to the last `orWhere`.

This meant that if a user had *any* open task then they were effectively granted permissions to action any task.

That is bad.


lennym:
I've updated this to add a second layer of validation to the critical ASRU statuses.

joefitter:
😬

lennym:
Yeah... it was a bad one. Hence my belt'n'braces solution.

PR #78, opened 2019-03-08
Add notifications hook

lvnhmd:


lennym:
Is this _actually_ WIP, because it looks legit to me.

lvnhmd:
@lennym the whole feature is WIP, hence WIP status

PR #84, opened 2019-04-04
save to s3 then send message with key

joefitter:


lennym:
DO NOT MERGE!

Blocking this until we have buckets in dev.

PR #89, opened 2019-04-11
Retrieve tasks by establishment

lvnhmd:


lennym:
The code looks fine, but you'll need to run `npm audit fix` to fix the vulnerabilities that the build is finding.

PR #91, opened 2019-04-15
Add comment placeholder hook

lvnhmd:


lib/hooks/comment/index.js
```diff
@@ -0,0 +1,4 @@
+
+module.exports = settings => () => {
+  return Promise.resolve();
```
lennym: Do you have a notion of what might appear in here?

I'd originally seen this and assumed it was a notify hook, which makes a lot of sense, but I can't imagine what standalone stuff should be in here.

lvnhmd: done

lennym:
Can you publish a new `taskflow` version (minor?) and bump in this PR pls?

PR #97, opened 2019-05-16
Add the model being modified to the metadata for tasks

wheelsandcogs:


lennym:
It's probably a lot easier, and not especially risky (because UUIDs) to just modify the default query to include deleted records.

It also means that tasks before the delete task will remain visible after a completed delete task, which is probably desirable too.

lib/decorators/metadata.js
```diff
@@ -1,28 +1,35 @@
...
   return c => {
-
     const promises = [
       c.data.establishmentId && cache.query(Establishment, c.data.establishmentId),
       c.data.subject && cache.query(Profile, c.data.subject),
       c.data.changedBy && cache.query(Profile, c.data.changedBy)
     ];
 
+    if (c.data.id && c.data.model) {
+      const modelName = c.data.model === 'pil' ? c.data.model.toUpperCase() : capitalize(c.data.model);
```
joefitter: not sure if this scales to _all_ the models - thinking `ProjectVersion` and `AsruEstablishment`

PR #103, opened 2019-06-07
updated applicant tasks to include subject and changedBy, also moved …

joefitter:
…logic for myTasks into standard format

### DIFF
--- a/lib/query-builders/applicant.js
+++ b/lib/query-builders/applicant.js
@@ -6,19 +6,33 @@ const {
   withdrawnByApplicant
 } = require('../flow/status');
 
+const isApplicant = profile => builder => builder
+  .whereJsonSupersetOf('data', { changedBy: profile.id })
+  .orWhereJsonSupersetOf('data', { subject: profile.id });
+
 module.exports = {
+
+  myTasks: (queryBuilder, profile) => {
+    queryBuilder
+      .where(isApplicant(profile))
+      .andWhere('status', returnedToApplicant.id);
+  },
+
   outstanding: (queryBuilder, profile) => {
-    queryBuilder.whereJsonSupersetOf('data', { subject: profile.id })
+    queryBuilder
+      .where(isApplicant(profile))
       .andWhere('status', returnedToApplicant.id);
   },
 
   inProgress: (queryBuilder, profile) => {
-    queryBuilder.whereJsonSupersetOf('data', { subject: profile.id })
+    queryBuilder
+      .where(isApplicant(profile))
       .whereNotIn('status', [autoResolved.id, resolved.id, rejected.id, withdrawnByApplicant.id, returnedToApplicant.id]);
   },
 
   completed: (queryBuilder, profile) => {
-    queryBuilder.whereJsonSupersetOf('data', { subject: profile.id })
+    queryBuilder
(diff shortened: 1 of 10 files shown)


lib/query-builders/applicant.js
```diff
@@ -6,19 +6,33 @@ const {
   withdrawnByApplicant
 } = require('../flow/status');
 
+const isApplicant = profile => builder => builder
```
lennym: This hurts my eyes. Can we wrap it in some curlies?

wheelsandcogs: #jode

lib/query-builders/inspector.js
```diff
@@ -24,9 +37,12 @@ module.exports = {
...
       inspectorRecommended.id,
-      inspectorRejected.id,
-      returnedToApplicant.id
-    ]);
+      inspectorRejected.id
+    ])
+      .orWhere(builder => {
+        builder.where('status', returnedToApplicant.id)
+          .whereJsonNotSupersetOf('data', { changedBy: profile.id });
+      });
```
lennym: I'm not sure this bit is correct? It seems like inspectors will see all tasks that are returned to applicants and _not_ their own i.e. everything that is returned to _someone else_.

I suspect `whereJsonNotSupersetOf` should be `whereJsonSupersetOf`.

lennym: Having said that, I don't think this is necessary at all. Inspectors seeing returned to applicant things is handled by the "applicant" query builder.

(unless this is intended to do something else that I don't understand)

lib/query-builders/applicant.js
joefitter: hahaha

joefitter: FINE.

lib/query-builders/inspector.js
joefitter: the latter, we're showing all returned to applicant tasks here (in progress) that are _not_ the inspector's tasks

joefitter: it may not be the right way of doing it, but without some kind of check the returned to applicant task when you are an inspector and the applicant, appears in my tasks, outstanding, _and_ in progress

lennym: Ahhh, the collapsed code threw me. I'd read it as being part of the `outstanding` query.

Carry on...

test/data/tasks.js
```diff
@@ -195,5 +195,18 @@ module.exports = query => query.insert([
       changedBy: user.id
     },
     status: 'autoresolved'
+  },
+  {
+    id: uuid(),
+    data: {
+      data: {
+        conditions: 'conditions update'
```
lennym: If you give this task a `name` then you don't need the custom assertion code that you wrote.

joefitter: I did think that, but then I thought it was unrealistic to change the "name" on an update-conditions task, I figured name made sense on the others as they were all profiles, but I was obviously overthinking it and should just use name :D

lennym: Yeah, the data doesn't matter for the tests, so it was just a way of identifying the tasks for asserting in a human-readable way (the alternative would have been listing uuids)

PR #111, opened 2019-06-17
Establishment ids are stored as ints

wheelsandcogs:
Establishments weren't returning any open tasks because the query was searching for a string id.

### DIFF
--- a/lib/router/model-tasks.js
+++ b/lib/router/model-tasks.js
@@ -8,7 +8,7 @@ module.exports = taskflow => {
   Case.db(taskflow.db);
 
   router.param('modelId', (req, res, next, modelId) => {
-    req.modelId = modelId;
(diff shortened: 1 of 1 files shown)


lib/router/model-tasks.js
```diff
@@ -8,7 +8,7 @@ module.exports = taskflow => {
   Case.db(taskflow.db);
 
   router.param('modelId', (req, res, next, modelId) => {
-    req.modelId = modelId;
+    req.modelId = /^\d+$/.test(modelId) ? parseInt(modelId) : modelId;
```
lennym: Always add a base to a `parseInt` call. Otherwise odd things might occur (they won't in _this_ case, but it's good bulletproof socks)

```suggestion
    req.modelId = /^\d+$/.test(modelId) ? parseInt(modelId, 10) : modelId;
```

PR #114, opened 2019-06-21
decorate task with licenceHolder if present in data

joefitter:


lennym:
Can you move this out into its own decorator rather than bundling it in this one? The reason being that it only applies to a very small subset of tasks, and as far as I can isn't required for rendering task lists, so we don't want to add an extra property to the responses for 100% of tasks when rendered in lists, when it's only needed for the task view screen of very specific tasks.

See https://github.com/UKHomeOffice/asl-workflow/blob/master/lib/api.js#L36 for example of non-list decorators.

lib/decorators/metadata.js
```diff
@@ -14,7 +14,7 @@ module.exports = settings => {
     ];
 
     return Promise.all(promises)
-      .then(([establishment, subject, changedBy]) => {
+      .then(([establishment, subject, changedBy, licenceHolder]) => {
```
lennym:
```suggestion
      .then(([establishment, subject, changedBy]) => {
```

PR #124, opened 2019-07-17
Autoresolve licensing PEL amendment

lvnhmd:


lib/hooks/create/establishment.js
```diff
@@ -26,6 +26,16 @@ module.exports = settings => {
               return model.setStatus(withLicensing.id);
             }
           });
+      case 'update':
+        return Promise.resolve()
+          .then(() => Profile.query().findById(changedBy))
+          .then(profile => {
+            if (profile.asruLicensing) {
```
lennym: This should be `profile.asruUser && profile.asruLicensing`. It's a tiny edge case, but it's real.

```suggestion
            if (profile.asruUser && profile.asruLicensing) {
```

lvnhmd: done

PR #127, opened 2019-07-29
Set the model data action to withdraw

lvnhmd:


lib/hooks/withdraw/index.js
```diff
@@ -0,0 +1,18 @@
...
+      return Promise.resolve();
+    }
+
+    model.data.action = 'withdraw';
+
+    return messager(model.data)
+      .then(changelogModel => model.patch({ id: changelogModel.modelId }));
+  };
+
+};
```
joefitter: This hook will be called for _any_ model being withdrawn, which will result in missing resolver errors for anything other than a project - we probably want to scope this hook so it's only called for projects

lvnhmd: Done

lennym: No it isn't. There's no code that checks that this is a grant action on a project. 

Unless I'm missing something, all that you have done is rename the file.

PR #130, opened 2019-08-08
Add isOpen flag to tasks

wheelsandcogs:


lib/decorators/metadata.js
```diff
@@ -23,7 +24,8 @@ module.exports = settings => {
             subject,
             profile: subject,
             changedBy
-          }
+          },
+          isOpen: !closed().includes(c.status)
```
lennym: It's a very minor point, but I'd move the call to `closed()` right up to the top of the function so it only calls once ever.

PR #133, opened 2019-08-12
New decorator to add deadline prop to project application tasks

wheelsandcogs:


lib/decorators/deadline.js
```diff
@@ -0,0 +1,34 @@
+const { get } = require('lodash');
+const moment = require('moment-business-time');
+const Case = require('@ukhomeoffice/taskflow/lib/models/case');
```
wheelsandcogs: Is there a better way to get access to the Case model? This seems to work without requiring any further connection setup, but should we be passing it in instead?

lib/decorators/deadline.js
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    const isAmendment = get(c, 'data.modelData.status') === 'active';
+
+    if (isAmendment) {
+      return c;
+    }
+
+    return Promise.resolve()
+      .then(() => Case.find(c.id))
```
wheelsandcogs: we re-fetch the case to get the activity log, which won't be available on `c` if a list of tasks is being decorated.

lib/decorators/deadline.js
lennym: There _should_ be a better way. This way is bad. 

Let me have a look.

lib/decorators/deadline.js
```diff
@@ -0,0 +1,34 @@
+const { get } = require('lodash');
+const moment = require('moment-business-time');
+const Case = require('@ukhomeoffice/taskflow/lib/models/case');
+
+module.exports = settings => {
+  return c => {
+    if (get(c, 'data.model') !== 'project') {
```
lennym: This should also check that the action is a `grant`

lennym:
It probably makes sense to do the "complete and correct" check here as well, so that we don't need to do the extra data lookups and traversals in the cases where it's not complete and correct, and so the deadline isn't relevant.

lib/decorators/deadline.js
lennym: Looks like there's already 4 other instances of this in the project, so if I wanted to have feelings about it then it's too late.

Crack on!

lennym:
This looks good, but I'm going to defer merging it because I want to get the deployment up today and could do with having as few other moving parts as possible.

wheelsandcogs:
👍

lennym:
Can we rip a load of code out of the task screens now?

PR #136, opened 2019-08-23
Add metrics endpoint to workflow

lennym:
Exposes an endpoint to fetch a list of tasks that have been resolved in a defined timeframe.

Exposes only the task type, and some basic metadata about the number of iterations to complete and the time to complete.

### DIFF
--- a/lib/router/metrics.js
+++ b/lib/router/metrics.js
@@ -28,7 +28,7 @@ module.exports = taskflow => {
         return cases.map(c => {
           let type = `${c.data.model}-${c.data.action}`;
           if (c.data.model === 'project' && c.data.action === 'grant') {
-            const isAmendment = c.data.modelData.status === 'active';
+            const isAmendment = c.data.modelData.status !== 'inactive';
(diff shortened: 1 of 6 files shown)


lib/router/metrics.js
```diff
@@ -0,0 +1,46 @@
...
+      })
+      // exclude ASRU TEST
+      .then(cases => {
+        return cases.filter(c => c.data.establishmentId !== 1502162);
+      })
+      .then(cases => {
+        return cases.map(c => {
+          let type = `${c.data.model}-${c.data.action}`;
+          if (c.data.model === 'project' && c.data.action === 'grant') {
+            const isAmendment = c.data.modelData.status === 'active';
```
joefitter: didn't we define an amendment as `status != inactive`?

lennym: Someone needs to tell https://github.com/UKHomeOffice/asl-workflow/blob/master/lib/decorators/deadline.js#L9

lennym: I've changed both.

PR #138, opened 2019-08-23
Use correct timestamps

lennym:
The updatedAt timestamp doesn't actually work on tasks - it's always the same as the createdAt - so use the resolved event to get the closed timestamp for metrics.


joefitter:
Ah! Yeah I remember @wheelsandcogs fixing that for the asl DB - you need to add an onUpdate hook to the base model I think

PR #139, opened 2019-08-27
Add hook for revoking a ppl

wheelsandcogs:
~Waiting on https://github.com/UKHomeOffice/asl-resolver/pull/99~

### DIFF
--- a/lib/hooks/create/project.js
+++ b/lib/hooks/create/project.js
@@ -47,7 +47,7 @@ module.exports = settings => {
               return model.setStatus(autoResolved.id);
             }
 
-            // Project revoked by inspector - needs licensing review
+            // Project revoked by other asru user - needs licensing review
             return model.setStatus(withLicensing.id);
           });
     }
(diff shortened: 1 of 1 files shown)


lib/hooks/create/project.js
```diff
@@ -26,9 +28,28 @@ module.exports = settings => {
...
+              // Project revoked by establishment user - needs inspector review
+              return model.setStatus(withInspectorate.id);
+            }
+
+            if (profile.asruLicensing) {
+              // Project revoked by licensing, does not need review
+              return model.setStatus(autoResolved.id);
+            }
+
+            // Project revoked by inspector - needs licensing review
```
joefitter: not sure this is true - there are other internal asru roles which aren't licensing or inspector.

wheelsandcogs: Yeah but don't we want an ASRU admin / ASRU basic revocation to go through licensing as well? Should I just correct the comment?

joefitter: that's what I was getting at :) it's just slightly misleading and not immediately obvious what's happening from a glance

wheelsandcogs: 👍

PR #141, opened 2019-08-28
added unit test

joefitter:


test/data/tasks.js
```diff
@@ -208,5 +208,22 @@ module.exports = query => query.insert([
...
+  // test for the case where the applicant is not
+  // `changedBy` or `subject`
+  {
+    id: uuid(),
+    data: {
+      data: {
+        name: 'Submitted by HOLC'
+      },
+      changedBy: holc.id,
+      // this _should_ be the user.id, but in some cases is not.
```
wheelsandcogs: what's `this` in this instance? changedBy or subject? can we stick the comment on the end of the line, add some spacing or word it to be less ambiguous?

joefitter: sure

PR #142, opened 2019-08-30
Feature/update task

joefitter:
Recall task  …
* added recalled-by-applicant status
* added discarded by applicant status
* establishment admin users can discard/recall any actionable task at their establishment
* recalled tasks to not show in asru task lists

### DIFF
--- a/lib/hooks/update/index.js
+++ b/lib/hooks/update/index.js
@@ -1,19 +1,21 @@
-const { get, set } = require('lodash');
+const { get, merge } = require('lodash');
 const { BadRequestError } = require('@asl/service/errors');
 
 const updateModel = (model) => {
   const payload = get(model, 'meta.payload', {});
-  const updates = get(model, 'data');
+
+  const data = get(model, 'data');
+  const updates = {};
 
   if (payload.data) {
-    set(updates, 'data', { ...updates.data, ...payload.data });
+    updates.data = payload.data;
   }
 
   if (payload.meta) {
-    set(updates, 'meta', { ...updates.meta, ...payload.meta });
+    updates.meta = payload.meta;
   }
 
-  return model.patch(updates);
+  return model.patch(merge(data, updates));
 };
 
 module.exports = settings => {
--- a/lib/hooks/update/index.js
+++ b/lib/hooks/update/index.js
@@ -1,0 +1,32 @@
+const { get, set } = require('lodash');
+const { BadRequestError } = require('@asl/service/errors');
+
+const updateModel = (model) => {
+  const payload = get(model, 'meta.payload', {});
+  const updates = get(model, 'data');
+
+  if (payload.data) {
+    set(updates, 'data', { ...updates.data, ...payload.data });
+  }
+
+  if (payload.meta) {
+    set(updates, 'meta', { ...updates.meta, ...payload.meta });
+  }
+
+  return model.patch(updates);
+};
+
+module.exports = settings => {
+  return model => {
+    const type = get(model, 'data.model');
+
+    switch (type) {
+      case 'place':
+      case 'establishment':
+      case 'pil':
+        return updateModel(model);
+    }
+
+    throw new BadRequestError();
+  };
+};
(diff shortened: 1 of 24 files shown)


joefitter:
@joefitter
Feature - update data  …
* added new `updated` status which updates task data and meta then autoforwards to `resubmitted`
* added hook to update data on status change
* disallow `updated` for certain models/actions

lib/hooks/update/index.js
```diff
@@ -0,0 +1,32 @@
+const { get, set } = require('lodash');
+const { BadRequestError } = require('@asl/service/errors');
+
+const updateModel = (model) => {
+  const payload = get(model, 'meta.payload', {});
+  const updates = get(model, 'data');
```
wheelsandcogs: this is a bit confusing here cause `updates` will be the existing model data at this point won't it?

What about something like 

```
const data = get(model, 'data');
const updates = {};

if (payload.data) {
  updates.data = payload.data;
}

if (payload.meta) {
  updates.meta = payload.meta;
}

return model.patch(merge(data, updates));
```

lib/query-builders/admin.js
```diff
@@ -20,18 +22,34 @@ module.exports = {
...
     queryBuilder
       .where(buildEstablishmentQuery(profile))
-      .whereNotIn('status', [autoResolved.id, resolved.id, rejected.id, withdrawnByApplicant.id, returnedToApplicant.id]);
+      .whereNotIn('status', [
+        autoResolved.id,
+        resolved.id,
+        rejected.id,
+        withdrawnByApplicant.id,
+        returnedToApplicant.id,
+        recalledByApplicant.id,
```
wheelsandcogs: Recalled tasks are hidden from ASRU right? 👍

lib/query-builders/applicant.js
```diff
@@ -18,24 +20,43 @@ module.exports = {
   myTasks: (queryBuilder, profile) => {
     queryBuilder
       .where(isApplicant(profile))
-      .andWhere('status', returnedToApplicant.id);
+      .whereIn('status', [
+        returnedToApplicant.id,
+        recalledByApplicant.id
+      ]);
```
wheelsandcogs: Applicant "myTasks" is a thing?

lib/query-builders/applicant.js
```diff
@@ -18,24 +20,43 @@ module.exports = {
...
   },
 
   outstanding: (queryBuilder, profile) => {
     queryBuilder
       .where(isApplicant(profile))
-      .andWhere('status', returnedToApplicant.id);
+      .whereIn('status', [
+        returnedToApplicant.id,
+        recalledByApplicant.id
```
wheelsandcogs: Looks like `myTasks` and `outstanding` return the same results? Should we bin off `myTasks` for applicants?

lib/query-builders/admin.js
wheelsandcogs: Ah hang on this isn't ASRU this is establishment admin. Probably still makes sense though.

lib/hooks/update/index.js
joefitter: sure, if you use the handy insert suggestion, then I just have to click the button :)

lib/query-builders/admin.js
joefitter: recalled are seen by admins, but not asru

lib/query-builders/applicant.js
joefitter: it is, but it's not used in the UI

lib/query-builders/applicant.js
joefitter: perhaps, but out of scope for this, as its there currently

joefitter: it would be useful for people who have access to lots of tasks, but only care about their ones

lib/hooks/update/index.js
joefitter: actually that won't work cos you chucked a merge in there, updating

PR #151, opened 2019-09-16
Replace whereNotIn queries with a whereIn

lennym:
The `whereNotIn` clause in a query can't hit the database indexes on the status column and so the queries run far slower than other task-list queries.

Replace the `whereNotIn` with a `whereIn` used in conjunction with a filter against a list of excluded statuses so that the indexes are used for the query.

### DIFF
--- a/lib/query-builders/admin.js
+++ b/lib/query-builders/admin.js
@@ -1,3 +1,6 @@
+const flow = require('../flow/status');
+const allStatuses = Object.values(flow).map(s => s.id);
+
 const {
   autoResolved,
   returnedToApplicant,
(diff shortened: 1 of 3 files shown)


lib/query-builders/admin.js
```diff
@@ -29,17 +32,18 @@ module.exports = {
   },
 
   inProgress: (queryBuilder, profile) => {
+    const nopes = [
```
joefitter: heh

lib/query-builders/admin.js
```diff
@@ -1,3 +1,6 @@
+const flow = require('../flow/status');
+const allStatuses = Object.values(flow).map(s => s.id);
```
wheelsandcogs: A `getAllStatusIds()` function in `lib/flow/index.js` could be useful.

PR #153, opened 2019-09-16
update data for project if action is update

joefitter:


lib/hooks/update/index.js
```diff
@@ -17,6 +17,14 @@ const updateModel = (model) => {
   return model.patch(merge(data, updates));
 };
 
+const updateProject = model => {
+  const action = get(model, 'data.action');
+  if (action === 'update') {
```
lennym:
```suggestion
  // only need to update on resubmit if action is to transfer PPL holder
  if (action === 'update') {
```

lib/hooks/edit-and-resubmit/index.js
```diff
@@ -17,6 +17,14 @@ const updateModel = (model) => {
   return model.patch(merge(data, updates));
 };
 
+const updateProject = model => {
+  const action = get(model, 'data.action');
+  if (action === 'update') {
```
lennym:
```suggestion
  // only need to update model data if task is to transfer PPL holder
  if (action === 'update') {
```

PR #155, opened 2019-09-20
Filter comments from activity log

lennym:
Removes comments based on the user role, the commenter's role and time relative to particular events. Users can always see comments made by people of the same "type" as them (i.e. ASRU or not-ASRU). For ASRU users viewing external comments they only become visible post-submission. For external users viewing ASRU comments they only become visible after an ASRU instigated status change.

Also moves the comment deletion code into here because it makes more sense.


lennym:
Tests coming...

wheelsandcogs:
Looks fine. Can we have some tests around comment deletion?

PR #158, opened 2019-09-23
Restrict tasks for external users to only the currently associated establishments

wheelsandcogs:


lennym:
I'm not super comfortable with the fact that this does a refactor of how the establishment queries are constructed but without any test coverage for users with multiple establishments.

It would be good to have at least one test case which ensures that a user with admin at two establishments can see a combined task list.

wheelsandcogs:
> I'm not super comfortable with the fact that this does a refactor of how the establishment queries are constructed but without any test coverage for users with multiple establishments.
> 
> It would be good to have at least one test case which ensures that a user with admin at two establishments can see a combined task list.

Done.

PR #165, opened 2019-10-16
Fix endorsement journey

lennym:
* "PPL Clock" should only start when an application arrives with the inspectorate
* Once endorsed an application should stay endorsed
* Amendments should not need to be endorsed
* Draft applications should not be marked as submitted until they have been endorsed

### DIFF
--- a/lib/hooks/create/project.js
+++ b/lib/hooks/create/project.js
@@ -1,12 +1,21 @@
 const { get } = require('lodash');
-const { autoResolved, withLicensing, awaitingEndorsement, endorsed } = require('../../flow/status');
+const { autoResolved, withLicensing, awaitingEndorsement, endorsed, withInspectorate } = require('../../flow/status');
 const { canEndorse } = require('../../util');
-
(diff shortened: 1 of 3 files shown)


lib/hooks/create/project.js
```diff
@@ -27,26 +39,35 @@ module.exports = settings => {
...
+              return model.setStatus(withInspectorate.id);
             }
           });
 
       case 'grant':
-        return messager({ ...model.data, action: 'submit-draft' })
-          .then(() => canEndorse(model))
-          .then(can => model.setStatus(can ? endorsed.id : awaitingEndorsement.id));
+        return Promise.resolve()
+          .then(needsEndorsing => {
```
joefitter: isn't this always `undefined`?

joefitter: also it doesn't look like it is used so will prob fail lint

lennym: It looks a lot like I did a half-arsed copy-paste there doesn't it?

PR #173, opened 2019-10-31
Auto-resolve licensing officer amendments to pils

wheelsandcogs:


lib/hooks/create/pil.js
```diff
@@ -30,6 +30,9 @@ module.exports = settings => {
         return model.setStatus(withLicensing.id);
       case 'grant':
         if (profile.asruUser && isAmendment) {
```
joefitter: so if an asru user _applies_ on behalf of somebody, that still goes to NTCO? The doesn't seem right

joefitter:
```suggestion
        if (profile.asruUser) {
```

PR #180, opened 2019-11-21
Transfer a PIL

wheelsandcogs:


lennym:
The establishment id that is created on this task is the id of the _receiving_ establishment, right?

wheelsandcogs:
> The establishment id that is created on this task is the id of the _receiving_ establishment, right?

Correct.

lennym:
Is the name of the outbound establishment stored/displayed anywhere?

There was some chat about showing that (and only that) as part of the task screen, but I can't see any code that does it.

wheelsandcogs:
The incoming pages PR shows the establishment name on task view even when it's not a transfer.

lennym:
Can we have some unit tests for this?

PR #185, opened 2019-11-28
Remove edit and resubmit as next step option for admins on pil transfer

wheelsandcogs:
The receiving admin on a PIL transfer cannot edit the request as the PIL is still owned by the original establishment until the task is approved. Remove the option to edit unless it's the applicant viewing the task.


lennym:
Could we get a unit test for this please?

PR #188, opened 2019-12-03
Feature - asru updating roles

joefitter:
* if role it holc, autoresolve
* if changedBy is licensing autoresolve
* if changedBy is asru send to licensing
* if changedBy is not asru send to inspectorate


wheelsandcogs:
Tests please.

joefitter:
:trollface:

PR #191, opened 2019-12-16
Added route to return open tasks for subject

joefitter:


lib/query-builders/index.js
```diff
@@ -1,11 +1,12 @@
 const { get, isUndefined } = require('lodash');
 const Case = require('@ukhomeoffice/taskflow/lib/models/case');
+const { resolved, rejected, withdrawnByApplicant, discardedByApplicant } = require('../flow/status');
```
lennym: There is a separate helper for getting closed/open tasks. Use that instead of re-enumerating here.

lennym: https://github.com/UKHomeOffice/asl-workflow/blob/master/lib/flow/index.js#L77-L84

lib/router/tasks.js
```diff
@@ -25,5 +25,14 @@ module.exports = (taskflow) => {
...
   });
 
+  router.get('/profile-tasks/:profileId', (req, res, next) => {
+    const profileId = req.params.profileId;
+    const send = taskflow.responder(req, res);
+    Promise.resolve()
+      .then(() => queryBuilder.getSubjectTasks(profileId))
+      .then(send)
+      .catch(next);
+  });
```
lennym: This should be a separate router - like https://github.com/UKHomeOffice/asl-workflow/blob/master/lib/router/model-tasks.js - rather than being piggy-backed onto the global task lists.

lennym:
Also, I believe there is another ticket which is similar to this, which is to prevent a person being removed from an establishment with open tasks. It might be worth adding an optional `establishmentId` parameter to this endpoint while you're here.

PR #193, opened 2019-12-30
Feature - billing

joefitter:
* Added model-history router which returns all tasks for a model filtered by status and action


lennym:
Based on my understanding of how things work currently, none of this is needed in order to deliver billing _without_ the ability to toggle billable-ness.

PR #196, opened 2020-01-10
Show establishment tasks in "My Tasks" tab

wheelsandcogs:
Establishment tasks don't send an `establishmentId` param on creation (just `id`) so it was not being copied to the task in the expected place for the queryBuilder to query on it.

### DIFF
--- a/lib/query-builders/index.js
+++ b/lib/query-builders/index.js
@@ -69,8 +69,6 @@ const buildQuery = (progress, profile, filters) => {
     }
   });
 
-  console.log(query.toString());
-
   return query;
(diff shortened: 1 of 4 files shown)


lib/query-builders/index.js
```diff
@@ -69,6 +69,8 @@ const buildQuery = (progress, profile, filters) => {
     }
   });
 
+  console.log(query.toString());
```
lennym: Remove this pls.

wheelsandcogs: doh

PR #198, opened 2020-01-15
Allow ASRU LOs to update the issue date of a project

wheelsandcogs:
Waiting on https://github.com/UKHomeOffice/asl-resolver/pull/123


test/data/default.js
```diff
@@ -78,19 +78,19 @@ module.exports = models => {
...
           asruUser: true,
           asruInspector: true
+        },
+        {
+          id: 'a942ffc7-e7ca-4d76-a001-0b5048a057d2',
+          firstName: 'Li Sen',
+          lastName: 'Xing',
+          email: 'lisenxing@example.com',
+          asruUser: true,
+          asruLicensing: true
```
wheelsandcogs: The id for inspectors and licensing were transposed in the seeds when compared with https://github.com/UKHomeOffice/asl-workflow/blob/master/test/data/profiles.js

wheelsandcogs: Same for the tests - could've just fixed them in profiles.js but I'd already done half of them by that point

lib/hooks/create/project.js
```diff
@@ -43,6 +44,17 @@ module.exports = settings => {
             }
           });
 
+      case 'update-issue-date':
+        return Promise.resolve()
+          .then(() => Profile.query().findById(changedBy))
+          .then(profile => {
+            if (!profile.asruUser || !profile.asruLicensing) {
+              throw new UnauthorisedError('Only ASRU licensing officers can change the granted date');
+            }
```
lennym: I'm not sure this check is actually good. There's a permissions check that looks for permissions on the task `project.updateIssueDate` upstream, and then this does a separate - potentially conflicting - check.

I think I'd prefer that if we were going to check permissions here (which I have no particular concern with as a concept) then it was via a permissions lookup rather than re-implementing the same check in a different way.

lib/hooks/create/project.js
```diff
@@ -43,6 +44,17 @@ module.exports = settings => {
             }
           });
 
+      case 'update-issue-date':
+        return Promise.resolve()
+          .then(() => Profile.query().findById(changedBy))
+          .then(profile => {
+            if (!profile.asruUser || !profile.asruLicensing) {
```
lennym:
```suggestion
            if (!profile.asruUser) {
```

PR #202, opened 2020-01-27
Add option for ASRU admins to discard open tasks

wheelsandcogs:


test/data/ids.js
```diff
@@ -1,17 +1,28 @@
...
-    grant: uuid()
+  model: {
+    pil: {
+      applied: uuid(),
+      rejected: uuid()
+    },
+    place: {
+      applied: uuid(),
+      resolved: uuid()
+    }
```
wheelsandcogs: This was really confusing when I was writing the tests - some ids referred to task ids and some referred to model ids, so I've tried to put a bit more structure in there. They're still not great but at least it's less confusing now.

lennym:
Looks good, but I'm going to deploy the in flight stuff before getting into this.

PR #214, opened 2020-02-10
Establishment revoke

wheelsandcogs:
Waiting on https://github.com/UKHomeOffice/asl-resolver/pull/131

### DIFF
--- a/lib/hooks/create/establishment.js
+++ b/lib/hooks/create/establishment.js
@@ -55,10 +55,10 @@ module.exports = settings => {
         return Promise.resolve()
           .then(() => Profile.query().findById(changedBy))
           .then(profile => {
-            if (!profile.asruUser || !profile.asruLicensing) {
-              throw new BadRequestError('Only ASRU licensing officers can revoke establishments');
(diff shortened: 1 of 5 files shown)


lib/hooks/create/establishment.js
```diff
@@ -51,6 +51,16 @@ module.exports = settings => {
       case 'grant':
         return model.setStatus(withInspectorate.id);
 
+      case 'revoke':
+        return Promise.resolve()
+          .then(() => Profile.query().findById(changedBy))
+          .then(profile => {
+            if (!profile.asruUser || !profile.asruLicensing) {
```
joefitter: I would _prefer_ to go for a `true && true` here - rather than a `false || false` - also can a user ever be asruLicensing and not asruUser? If not then we could just do an `if (profile.asruLicensing)...`

wheelsandcogs: There was a time when we just removed `asruUser` and remembered `asruLicensing` if an LO was removed from ASRU, but that should no longer be the case. This is just for belt and braces.

joefitter: 👍

PR #220, opened 2020-02-25
PPL - change establishment

joefitter:
* Added to endorse hook for projects - if transfer then update task establishmentId
* Added transfer to create hook to follow same path as grant tasks

### DIFF
--- a/lib/hooks/create/project.js
+++ b/lib/hooks/create/project.js
@@ -10,12 +10,15 @@ module.exports = settings => {
   const messager = Messager(settings);
 
   const needsEndorsement = (model) => {
+    const isEndorsed = get(model, 'data.meta.authority', 'no').toLowerCase() === 'yes';
+    const isTransfer = get(model, 'data.action') === 'transfer';
+    const isAmendment = get(model, 'data.modelData.status', 'inactive') === 'active';
     // don't require re-endorsing
-    if (get(model, 'data.meta.authority', 'no').toLowerCase() === 'yes') {
+    if (isEndorsed) {
       return false;
     }
     // don't require endorsement on amendments
-    if (get(model, 'data.modelData.status', 'inactive') === 'active') {
+    if (!isTransfer && isAmendment) {
       return false;
     }
     return true;
--- a/lib/hooks/create/project.js
+++ b/lib/hooks/create/project.js
@@ -56,6 +56,7 @@ module.exports = settings => {
           });
 
       case 'grant':
+      case 'transfer':
         return Promise.resolve()
           .then(() => {
             if (!needsEndorsement(model)) {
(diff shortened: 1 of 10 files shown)


lib/hooks/endorsed/index.js
```diff
@@ -9,10 +9,16 @@ const Messager = require('../../messager');
 
 module.exports = settings => {
 
-  const messager = Messager(settings);
+  const messager = settings.StubMessager ? settings.StubMessager : Messager(settings);
```
joefitter: sorrybox.gif

lib/hooks/create/project.js
```diff
@@ -10,12 +10,15 @@ module.exports = settings => {
   const messager = Messager(settings);
 
   const needsEndorsement = (model) => {
+    const isEndorsed = get(model, 'data.meta.authority', 'no').toLowerCase() === 'yes';
+    const isTransfer = get(model, 'data.action') === 'transfer';
+    const isAmendment = get(model, 'data.modelData.status', 'inactive') === 'active';
```
lennym:
```suggestion
    const isAmendment = get(model, 'data.modelData.status', 'inactive') === 'active';
```

lennym: Where we have the default values in the `get` here can we remove them because they add precisely no value, but cause confusion?

(Yes, I know it was like that before)

lib/hooks/endorsed/project.js
```diff
@@ -0,0 +1,54 @@
+const { get, merge } = require('lodash');
+const { BadRequestError } = require('@asl/service/errors');
+const { awaitingEndorsement, withInspectorate } = require('../../flow/status');
+const Messager = require('../../messager');
+
+module.exports = settings => model => {
+  const messager = settings.StubMessager ? settings.StubMessager : Messager(settings);
```
lennym: It feels like this is something that should be done inside the constructor given we pass the settings in, and not leaked out here.

lennym: Also creating the messager instance should probably occur outside the inner function.

lib/hooks/endorsed/index.js
```diff
@@ -1,32 +1,21 @@
...
-        return Promise.resolve()
-          .then(() => {
-            return messager({ ...model.data, action: 'submit-draft' });
-          })
-          .then(() => {
-            const data = merge(model.data, { meta: { authority: 'Yes' } });
-            return model.patch(data);
-          })
-          .then(() => model.setStatus(withInspectorate.id));
+        return project(settings)(model);
```
lennym: Again, can we pull the `project(settings)` bit up into the outer function so we're not re-initialising all the things per invocation?

joefitter: ah yeah  - good spot

PR #230, opened 2020-03-02
Bugfix - fork version on recall

joefitter:
We were only forking recalled/returned versions on grant tasks - updated to include transfer


lennym:
You got a unit test for this?

joefitter:
yep

PR #231, opened 2020-03-03
Bugfix - dont update id post resolve

joefitter:
* We were using the task id to inform the ui of the new version when forking, update the data.versionId instead
* Don't update id if there is already an id set


joefitter:
This is a WIP as it is a breaking change until https://github.com/UKHomeOffice/asl-pages/pull/719

PR #234, opened 2020-03-09
Feature - ppl transfer

joefitter:
* Updated unique check to update project grant status to resubmitted if already present, point to existing record rather than creating a new task
* Added update-action hook, this updates the action type of a grant task if it is detected to be a transfer

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -36,7 +36,8 @@ const hooks = {
   editAndResubmit: require('./hooks/edit-and-resubmit'),
   discard: require('./hooks/discard'),
   recall: require('./hooks/recall'),
-  endorsed: require('./hooks/endorsed')
+  endorsed: require('./hooks/endorsed'),
+  updateAction: require('./hooks/update-action')
 };
 
 module.exports = settings => {
@@ -61,7 +62,9 @@ module.exports = settings => {
   flow.hook('pre-create', hooks.unique(settings));
 
   flow.hook('create', hooks.meta(settings));
+  flow.hook('create', hooks.updateAction(settings));
   flow.hook('create', hooks.create(settings));
+  flow.hook(`status:*:${resubmitted.id}`, hooks.updateAction(settings));
   flow.hook(`status:*:${resubmitted.id}`, hooks.create(settings));
 
   flow.hook(`status:*:${recalledByApplicant.id}`, hooks.clearDeadline(settings));
--- a/lib/hooks/create/unique.js
+++ b/lib/hooks/create/unique.js
@@ -37,7 +37,7 @@ module.exports = settings => {
         if (results && results.length) {
           if (type === 'project' && action === 'grant') {
             const id = results[0].id;
-            return Task.find(id)
+            return model.find(id)
               .then(task => {
                 const status = task.toJSON().status;
                 if (editable().includes(status)) {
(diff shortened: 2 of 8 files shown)


lib/api.js
```diff
@@ -61,7 +62,9 @@ module.exports = settings => {
   flow.hook('pre-create', hooks.unique(settings));
 
   flow.hook('create', hooks.meta(settings));
+  flow.hook('create', hooks.updateAction(settings));
```
lennym: This is surely unnecessary by definition? If we're creating a new task then it can't be necessary to also update an existing task?

lib/hooks/create/unique.js
```diff
@@ -34,6 +35,12 @@ module.exports = settings => {
       .where('status', '!=', 'new')
       .then(results => {
         if (results && results.length) {
+          if (type === 'project' && action === 'grant') {
```
lennym: Might want to check if the open task is in an editable state here.

lib/api.js
lennym: I'm wrong here I think, right?

This does need to be here because the task is _always_ a `type: 'grant'` until it is modified here?

test/helpers/workflow.js
```diff
@@ -31,10 +31,10 @@ module.exports = {
     return knex.destroy();
   },
 
-  resetDBs: () => {
+  resetDBs: ({ keepalive = false } = {}) => {
     return Promise.resolve()
       .then(() => tfdb.reset())
-      .then(() => aslDb(settings.db).init(fixtures.default));
+      .then(() => aslDb(settings.db).init(fixtures.default, keepalive));
```
lennym: What is this for?

lennym: It keeps the database connection open so that we can query the models in the tests, right?

In which case we probably want to make sure we're closing the connections in the `afterEach`

lib/hooks/update-action/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+      // not a transfer - continue.
+      if (!transferToEstablishmentId) {
+        return Promise.resolve();
+      }
+
+      const transferToEstablishment = await Establishment.query().findById(transferToEstablishmentId);
+
+      if (!transferToEstablishment) {
+        throw new NotFoundError();
```
lennym: What happens to the task or project/version if one of these errors is thrown?

lib/api.js
joefitter: yes

lib/hooks/create/unique.js
joefitter: good shout

test/helpers/workflow.js
joefitter: yes, and good shout

lib/hooks/update-action/index.js
joefitter: god knows

lib/hooks/create/unique.js
```diff
@@ -34,6 +35,17 @@ module.exports = settings => {
...
+            const id = results[0].id;
+            return Task.find(id)
```
lennym: See https://github.com/UKHomeOffice/taskflow/pull/48

```suggestion
            return model.find(id)
```

lennym: Note: don't do this until that PR has been merged and published.

lib/hooks/update-action/index.js
lennym: It works fine :-)

joefitter: clever girl

PR #239, opened 2020-03-11
Prevent external users actioning ASRU tasks

lennym:
Currently licence holders and admins can discard ASRU initiated tasks, this should not be possible.

Remove all possible next steps for a user if they are not an ASRU user and the task was ASRU initiated. Additionally remove the tasks from their outstanding task lists if they are ASRU initiated.

### DIFF
--- a/lib/flow/get-next-steps-for-user.js
+++ b/lib/flow/get-next-steps-for-user.js
@@ -3,20 +3,20 @@ const flow = require('./index');
 const { updated, recalledByApplicant, discardedByApplicant, discardedByAsru } = require('./status');
 const queryBuilder = require('../query-builders');
 
-const userCreatedThisTask = (c, profile) => {
-  if (!c.data.changedBy) {
+const userCreatedThisTask = (task, profile) => {
+  if (!task.data.changedBy) {
     return false;
   }
 
-  if (typeof c.data.changedBy === 'string') {
-    return c.data.changedBy === profile.id;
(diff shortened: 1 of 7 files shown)


lib/flow/get-next-steps-for-user.js
```diff
@@ -25,37 +25,41 @@ const userIsAsruAdmin = profile => {
...
+module.exports = (task, profile) => {
+  if (flow.autoForwards(task.status)) {
+    return Promise.resolve(flow.getNextSteps(task));
   }
 
   let nextSteps = [];
 
+  if (task.data.initiatedByAsru && !profile.asruUser) {
+    return Promise.resolve([]);
+  }
```
lennym: Adding this block ☝️ is the only bit of this file change that _isn't_ `c => task`

lib/query-builders/admin.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
   outstanding: (queryBuilder, profile) => {
     queryBuilder
       .where(buildEstablishmentQuery(profile))
+      .whereJsonSupersetOf('data', { initiatedByAsru: false })
```
wheelsandcogs: is this going to potentially hide old tasks that do not have the `initiatedByAsru` prop?

lennym: Yes, it absolutely is.

I'll see what impact that has.

joefitter: do you want to hold off merging this?

PR #241, opened 2020-03-13
bugfix - activity log

joefitter:
* only include status change events and deadline updates
~* if multiple events happened within 5 secs, group taking the action from the first event, and the status from the last~
* check the status change sent in the request, if the end status of the activity log item doesn't match then group with the next item and show the status sent (action) and the end status (status)

### DIFF
--- a/lib/decorators/activity-log.js
+++ b/lib/decorators/activity-log.js
@@ -1,17 +1,21 @@
-const moment = require('moment');
+const { get } = require('lodash');
 const Cacheable = require('./cacheable');
 const { autoResolved } = require('../flow/status');
 
-function getStatusFromEvent(event) {
-  return event.split(':').pop();
+function getItemStatus(item) {
+  return item.eventName.split(':').pop();
+}
+
+function getStatusChange(item) {
+  return get(item, 'event.meta.payload.status');
+}
+
+function isDeadlineExtension(item) {
+  return !!get(item, 'event.meta.payload.data.extended');
 }
 
 function filterStatusChanges(item) {
-  return item.eventName.split(':')[0] === 'status';
-}
-
-function isThenish(date, then) {
-  return moment(date).isBetween(moment(then), moment(then).add(5, 'seconds'));
+  return item.eventName.split(':')[0] === 'status' || isDeadlineExtension(item);
 }
 
 module.exports = settings => {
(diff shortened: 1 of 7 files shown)


lennym:
We'll need to preserve deadline extension events as well, not _just_ status changes.

lib/decorators/activity-log.js
```diff
@@ -21,11 +39,32 @@ module.exports = settings => {
...
+            ];
+          }
+        }, []);
+
+        activityLog = activityLog.map(log => {
           log.changedBy = logChangedBys.find(profile => profile && profile.id === log.changedBy) || {};
           delete log.event.meta.user.access_token;
           return log;
-        });
+        }).reverse();
```
lennym: Might be worth making this an explicit `sortBy`, not necessarily because there's any reason to believe it's not correctly sorted, but because it makes it more obvious within the context of this file what the expected sort order is (and less potentially sensitive to breaking as a result of changes elsewhere).

test/unit/decorators/filter-comments.js
```diff
@@ -1,56 +1,29 @@
...
-const assertComments = (task, user, expected) => {
-  const result = decorator()(task.model, user);
-  const comments = result.activityLog.filter(a => a.eventName === 'comment');
+const assertComments = async (task, user, expected) => {
+  const settings = {
+    models: {
+      Profile: {
+        queryWithDeleted: sinon.stub().returns({
+          findById: sinon.stub().returns({
+            then: sinon.stub().returns([])
```
lennym: YIKES!

lib/decorators/activity-log.js
```diff
@@ -1,6 +1,24 @@
...
+  return get(item, 'event.meta.payload.status');
+}
+
+function isDeadlineExtension(item) {
+  return !!get(item, 'event.meta.payload.data.extended');
+}
+
+function filterStatusChanges(item) {
+  const [type, , status] = item.eventName.split(':');
+  return (type === 'status' && status !== 'resubmitted') || isDeadlineExtension(item);
```
lennym: What's the logic behind removing `resubmitted`?

joefitter: resubmitted autoforwards, but doesn't have a meta.payload.status - we display the thing that it autoforwards too eg with licensing. If we leave it in there are 2 logs for the same event

PR #243, opened 2020-03-19
Feature - transfer meta

joefitter:
* added to and from establishment meta to task if transfer

### DIFF
--- a/lib/hooks/update-action/index.js
+++ b/lib/hooks/update-action/index.js
@@ -1,4 +1,4 @@
-const { get } = require('lodash');
+const { get, pick } = require('lodash');
 const { NotFoundError, BadRequestError } = require('@asl/service/errors');
 
 module.exports = settings => {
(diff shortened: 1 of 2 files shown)


lib/hooks/update-action/index.js
```diff
@@ -39,7 +39,16 @@ module.exports = settings => {
...
       }
 
-      return model.patch({ action: 'transfer', data: { establishmentId: transferToEstablishmentId } });
+      const transferFromEstablishment = await Establishment.query().findById(project.establishmentId);
+
+      const meta = {
+        establishment: {
+          to: pick(transferToEstablishment, 'id', 'name'),
+          from: pick(transferFromEstablishment, 'id', 'name')
+        }
```
wheelsandcogs: This looks familiar :D

joefitter: hehe turns out it's a hell of a lot easier to attach the meta at the point of creating the transfer than trying to work it out depending on state in the UI :D

PR #247, opened 2020-03-23
Feature - PIL review

joefitter:
* added create hook for review - autoresolve if ntco, set to awaitingEndorsement if not
* updated ntco query builder to include review pils
* added pil review hook to resolve task

### DIFF
--- a/lib/flow/status.js
+++ b/lib/flow/status.js
@@ -14,7 +14,7 @@ function isLicensing(task) {
 }
 
 function isNtcoReview(task) {
-  const isReview = get(task, 'data.action') === 'review';
+  const isReview = get(task, 'data.action') === 'review' && get(task, 'data.model') === 'pil';
   return isReview && canEndorse(task);
 }
 
--- a/lib/flow/status.js
+++ b/lib/flow/status.js
@@ -13,6 +13,11 @@ function isLicensing(task) {
   return isAsruUser(task) && get(task, 'meta.user.profile.asruLicensing', false);
 }
 
+function isNtcoReview(task) {
+  const isReview = get(task, 'data.action') === 'review';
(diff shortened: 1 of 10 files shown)


lib/flow/status.js
```diff
@@ -13,6 +13,11 @@ function isLicensing(task) {
   return isAsruUser(task) && get(task, 'meta.user.profile.asruLicensing', false);
 }
 
+function isNtcoReview(task) {
+  const isReview = get(task, 'data.action') === 'review';
```
lennym:
```suggestion
  const isReview = get(task, 'data.action') === 'review' && get(task, 'data.model') === 'pil';
```

lib/query-builders/ntco.js
```diff
@@ -28,6 +28,7 @@ module.exports = {
       .where(builder => {
         builder.whereJsonSupersetOf('data', { model: 'pil', action: 'grant' });
         builder.orWhereJsonSupersetOf('data', { model: 'pil', action: 'transfer' });
+        builder.orWhereJsonSupersetOf('data', { model: 'pil', action: 'review' });
```
lennym: Is there any kind of `pil` task that might have one of the specified statuses that we _wouldn't_ want in this query?

If not then probably we should consolidate these into a single `{ model: 'pil' }` query for efficiency.

joefitter: revoke

lennym: I'd still probably get rid of the action criteria and reduce this to a single clause.

It'll make the query faster, and task-list queries need all the help they can get, and on the outside chance that a revocation ended up in a `awaiting-endorsement` state then it's possibly still the right thing to include it.

PR #249, opened 2020-03-26
Bugfix - ASRU can be applicant

joefitter:
* don't blacklist asru tasks if user is asru


test/integration/task-lists/establishment-admin.js
```diff
@@ -34,7 +34,6 @@ describe('Establishment Admin', () => {
         'place update returned',
         'Submitted by HOLC',
         'recalled ppl',
-        'conditions update',
```
lennym: Oh yeah, this test _was_ bollocksed wasn't it.

PR #253, opened 2020-04-01
Provide default db password for running seeds and migrations in dev

wheelsandcogs:


config.js
```diff
@@ -18,14 +18,14 @@ module.exports = {
...
-    password: process.env.DATABASE_PASSWORD,
+    password: process.env.DATABASE_PASSWORD || 'test-password',
     port: process.env.DATABASE_PORT,
     user: process.env.DATABASE_USERNAME || 'postgres'
   },
   db: {
     database: process.env.ASL_DATABASE_NAME || 'asl',
     host: process.env.ASL_DATABASE_HOST,
-    password: process.env.ASL_DATABASE_PASSWORD,
+    password: process.env.ASL_DATABASE_PASSWORD || 'test-password',
```
lennym, line 28: Does anything _not_ work if these are not specified?

wheelsandcogs: `npm run migrate` seems to rely on them being in `config.js` and `npm run seed` needs them in `knexfile.js`

wheelsandcogs: both seem to ignore the `.env`

PR #257, opened 2020-04-16
Filter out tasks at blocked establishments when returning tasklist

wheelsandcogs:


lennym:
Can you add a test for the filter please?

wheelsandcogs:
done.

PR #259, opened 2020-04-17
Bugfix - resolve pil grant if asru

joefitter:
* previously this would autoresolve and not leave a task behind


lennym:
Is this going to break the content you just fixed?

joefitter:
no, I tested it - we have one for resolved. It does mean we have a superfluous success message now

lennym:
Nice. Apart from the "superfluous" bit, obvs.

joefitter:
🤷

PR #260, opened 2020-04-17
NTCO should not have endorse options for own PIL

wheelsandcogs:


lennym:
You got a unit test for this?

lennym:
Does it also filter the tasks out of their task lists? It probably should.

PR #262, opened 2020-04-30
Feature - continuation as structured data

joefitter:
* Added create hook for project grant/transfer to add continuation data to task if found in licence
* Added decorator to add continuation to task if undefined on project grant/transfer (in flight amendments)

### DIFF
--- a/lib/decorators/continuation.js
+++ b/lib/decorators/continuation.js
@@ -1,11 +1,8 @@
-const Cacheable = require('./cacheable');
 const { ref } = require('objection');
 const { get, isUndefined } = require('lodash');
 
 module.exports = settings => {
-
-  const { Project } = settings.models;
-  const cache = Cacheable();
+  const { ProjectVersion } = settings.models;
 
   return c => {
     const model = get(c, 'data.model');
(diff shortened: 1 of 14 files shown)


lib/decorators/continuation.js
```diff
@@ -0,0 +1,60 @@
...
+            .first()
+            .select(
+              ref('data:transfer-expiring')
+                .castBool()
+                .as('expiring'),
+              ref('data:project-continuation')
+                .castJson()
+                .as('continuation')
+            );
+        })
```
lennym, line 42: Do we not have the version id on the task, so can query it directly?

lib/hooks/create/project.js
```diff
@@ -94,6 +117,7 @@ module.exports = settings => {
       case 'grant':
       case 'transfer':
         return Promise.resolve()
+          .then(() => markContinuation(model))
```
lennym, line 120: Does this run on resubmission as well creation?

It will need to to ensure that any changes in the version are reflected on the task.

lib/decorators/continuation.js
joefitter: we do, I just wasn't sure whether to trust it is always kept up to date :)

lib/hooks/create/project.js
joefitter: yes, resubmissions too

test/unit/decorators/filter-comments.js
```diff
@@ -25,7 +25,7 @@ const assertComments = async (task, user, expected) => {
   const result = await decorator(settings)(task.model, user);
   const comments = result.comments;
   assert.equal(comments.length, expected.length);
-  assert.deepEqual(comments.map(c => c.comment), expected);
+  assert.deepEqual(comments.map(c => c.comment).sort(), expected.sort());
```
lennym, line 28: This _shouldn't_ be necessary (and is probably bad) because comment ordering is important.

The history helper should be setting timestamps in a way that respects ordering, and the comments should be returned ordered by timestamp. If they're not then the underlying cause should be fixed rather than slapping a sort on them.
```suggestion
  assert.deepEqual(comments.map(c => c.comment), expected);
```

joefitter: yeah, bit of a lane swerve as they weren't actually running before. But I'll investigate why they are failing seeing as I made them run

lib/decorators/continuation.js
lennym: Looking at the task read router and what is used to render the "View submitted version" it looks safe to assume the version id is correct.

PR #265, opened 2020-04-30
Allow an establishment grant to be opened alongside other establishment tasks

lennym:
If there's an open amendment to the establishment details still allow the establishment to apply to be granted. This is for consistency with other PEL changes (places, roles, etc) that do not block the application.


lennym:
I'm going to ask _myself_ for a unit test on this one actually.

PR #267, opened 2020-05-11
Autoresolve role updates on places

wheelsandcogs:
~Needs
https://github.com/UKHomeOffice/asl-schema/pull/313
https://github.com/UKHomeOffice/asl-resolver/pull/155~

### DIFF
--- a/lib/hooks/create/place.js
+++ b/lib/hooks/create/place.js
@@ -22,7 +22,7 @@ const onlyRolesChanged = model => {
   const currentModel = get(model, 'data.modelData');
   const data = get(model, 'data.data');
 
-  const oldValues = omit(currentModel, ['id', 'deleted', 'createdAt', 'updatedAt', 'nacwoId', 'migratedId']);
+  const oldValues = omit(currentModel, ['id', 'deleted', 'createdAt', 'updatedAt', 'nacwoId', 'migratedId', 'roles']);
   const newValues = omit(data, ['roles']);
   const changes = diff(oldValues, newValues, { normalize });
 
--- a/lib/hooks/create/place.js
+++ b/lib/hooks/create/place.js
@@ -10,9 +10,9 @@ const normalize = (path, key, lhs, rhs) => {
     return [lhs, rhs];
   }
 
-  // handle empty db values vs empty form fields
-  if (lhs === null) {
-    return ['', rhs];
+  // handle different falsy values (null, empty string, undefined)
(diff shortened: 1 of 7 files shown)


lib/hooks/create/place.js
```diff
@@ -1,9 +1,39 @@
...
+
+const onlyRolesChanged = model => {
+  const currentModel = get(model, 'data.modelData');
+  const data = get(model, 'data.data');
+
+  const oldValues = omit(currentModel, ['id', 'deleted', 'createdAt', 'updatedAt', 'nacwoId', 'migratedId']);
+  const newValues = omit(data, ['roles']);
+  const changes = diff(oldValues, newValues, { normalize });
+
+  return !changes;
```
lennym, line 29: whoa.gif

wheelsandcogs: I didn't want to hard code the fields that we compare in case we add new fields to the place table.

lib/hooks/create/place.js
```diff
@@ -1,9 +1,39 @@
...
+
+  return false;
+};
+
+const onlyRolesChanged = model => {
+  const currentModel = get(model, 'data.modelData');
+  const data = get(model, 'data.data');
+
+  const oldValues = omit(currentModel, ['id', 'deleted', 'createdAt', 'updatedAt', 'nacwoId', 'migratedId']);
+  const newValues = omit(data, ['roles']);
```
lennym, line 26: We're _not_ omitting `roles` from the old value. Does that not mean that if the old value has roles then this will always detect changes.

wheelsandcogs: The old values come from the `task.data.modelData` and now that roles come from a join, it doesn't end up in the task.

Thinking about it, we might want to change this to look up the current values of the place from the db to compare against, although I don't think you can have multiple inflight tasks for a place so it shouldn't be possible to change the data elsewhere between this task being created and resolved?

lennym: Unless the answer is that the current model doesn't have roles attached, in which case how does the task diff table render roles?

lennym: Given there's a number of tickets coming to fix issues with the diff tables _not_ using the `modelData` to render the "before" state (and hence ending up weird once resolved) I think the `modelData` _should_ include the roles, and therefore should also be omitted here.

In short - omit `roles`

PR #268, opened 2020-05-28
Feature - model data

joefitter:
* Added hook to attach model data per type
* ensure db connections are closed after tests

### DIFF
--- a/test/helpers/workflow.js
+++ b/test/helpers/workflow.js
@@ -8,6 +8,7 @@ const settings = require('./database-settings');
 
 let knex = {};
 let tfdb = {};
+let workflow;
 
 module.exports = {
   create: (options = {}) => {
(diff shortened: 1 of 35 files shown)


test/helpers/workflow.js
```diff
@@ -28,13 +29,15 @@ module.exports = {
   },
 
   destroy: () => {
-    return knex.destroy();
+    return Promise.resolve()
+      .then(() => knex.destroy())
+      .then(() => workflow.flow.db.destroy());
```
lennym, line 34: I'm guessing this line was the crux of the database connection issue?

lennym: I'm not certain this will actually fix the issue reliably given the way it's currently scoped though.

The `workflow` and `knex` variables are scoped to the top level of the module, and so every call to create will reassign them, creating new connections each time, but the `destroy` call will only tear down the most recently created.

joefitter: ORLY yeah this "fixes" the issue, in that it actually runs them all

PR #269, opened 2020-05-28
Scope test db teardown methods to an instance

lennym:
Previously a single global reference to an instance was maintained and teardown methods pointed only to this insitance, and would leave behind others.

By scoping the teardown and seed methods to each instance then we can ensure that all instances are correctly removed.

### DIFF
--- a/test/integration/hooks/model-data.js
+++ b/test/integration/hooks/model-data.js
@@ -14,12 +14,12 @@ describe('Model data hook', () => {
 
   beforeEach(() => {
     return Promise.resolve()
-      .then(() => workflowHelper.resetDBs())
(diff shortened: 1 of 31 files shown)


test/integration/hooks/model-data.js
```diff
@@ -115,7 +115,7 @@ describe('Model data hook', () => {
         .then(response => response.body.data)
         .then(task => {
           assert.equal(task.data.modelData.name, 'Has assigned roles');
-          const role = task.data.modelData.roles[0]
+          const role = task.data.modelData.roles[0];
```
lennym, line 118: You're welcome.

joefitter: xox

PR #278, opened 2020-06-17
Fix query builder for generic ASRU user

lennym:
It was not specifying the id correctly for `discarded-by-asru`.

Refactor task tests so that they better assert on tasks which are not present in a result set.

### DIFF
--- a/test/helpers/assert-tasks.js
+++ b/test/helpers/assert-tasks.js
@@ -1,7 +1,14 @@
 const assert = require('assert');
+const { difference } = require('lodash');
+const seeds = require('../data/tasks');
 
 module.exports = (expected, actual) => {
-  assert.equal(actual.length, expected.length, `Expected ${expected.length} records. Received ${actual.length}`);
   const names = actual.map(task => task.data.data.name);
-  names.forEach(name => assert(expected.includes(name), `Expected [${expected.join(', ')}] to include "${name}"`));
+  const all = seeds.map(task => task.data.data.name);
+  const nopes = difference(all, names);
+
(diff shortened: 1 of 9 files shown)


test/helpers/assert-tasks.js
```diff
@@ -1,7 +1,14 @@
...
 
 module.exports = (expected, actual) => {
-  assert.equal(actual.length, expected.length, `Expected ${expected.length} records. Received ${actual.length}`);
   const names = actual.map(task => task.data.data.name);
-  names.forEach(name => assert(expected.includes(name), `Expected [${expected.join(', ')}] to include "${name}"`));
+  const all = seeds.map(task => task.data.data.name);
+  const nopes = difference(all, names);
+
+  names.forEach(name => assert(expected.includes(name), `[${expected.join(', ')}] should include "${name}"`));
+  nopes.forEach(name => assert(!names.includes(name), `[${actual.join(', ')}] should not include "${name}"`));
```
lennym, line 11: This now calculates a list of all the tasks it should _not_ expect and checks they are not present, which is a better output than simply comparing the length.

test/helpers/taskflow-db.js
```diff
@@ -13,7 +13,7 @@ module.exports = knex => {
       }),
     seed: (tasks = []) => Promise.resolve()
       .then(() => {
-        return seeds(Task.query(knex));
+        return Task.query(knex).insert(seeds);
```
lennym, line 16: This has moved from the seed file so it can be imported for the raw data.

wheelsandcogs:
Nice work.

PR #281, opened 2020-06-22
Optimise activity log data payloads

lennym:
Don't eager load ASRU establishment assignments onto profiles by default. These are quite large objects, which use a lot of space on the profile and are saved on every activity, where tehy're onyl actualyl used for loading task lists.

Instead only load ASRU users establishment assignments when building a task list.

Also remove access token from the user, since it isn't used downstream and while it represents only a very mild security risk since it is short-lived, it has no business being saved to data.


joefitter:
"where tehy're onyl actualyl"
- @lennym, 2020

PR #282, opened 2020-06-22
Feature - pil training on submission

joefitter:
* Added new hook to attach training an exemptions to task on submission and resubmission

### DIFF
--- a/lib/hooks/submission-meta/index.js
+++ b/lib/hooks/submission-meta/index.js
@@ -1,0 +1,26 @@
+const { get } = require('lodash');
+
+module.exports = settings => {
+  const { Certificate, Exemption } = settings.models;
+
+  return model => {
+    const type = get(model, 'data.model');
+    const action = get(model, 'data.action');
+
+    if (type !== 'pil') {
+      return Promise.resolve();
+    }
+
+    if (action !== 'grant' && action !== 'transfer') {
+      return Promise.resolve();
+    }
+
+    const profileId = get(model, 'data.data.profileId');
+
+    return Promise.all([
+      Certificate.query().where({ profileId }),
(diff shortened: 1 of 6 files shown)


lib/hooks/submission-meta/index.js
```diff
@@ -0,0 +1,26 @@
...
+
+  return model => {
+    const type = get(model, 'data.model');
+    const action = get(model, 'data.action');
+
+    if (type !== 'pil') {
+      return Promise.resolve();
+    }
+
+    if (action !== 'grant' && action !== 'transfer') {
```
lennym, line 14: What are the things we're trying to ignore here? All I can think of is reviews, and it seems like perhaps they should have training attached too.

test/data/default.js
```diff
@@ -86,7 +86,23 @@ module.exports = models => {
...
+              modules: [
+                {
+                  module: 'E1'
+                }
+              ]
+            }
+          ],
+          exemptions: [
+            {
+              id: ids.model.exemption.colinJackson,
```
lennym, line 102: 🤣

lib/hooks/submission-meta/index.js
joefitter: yeah, I thought that, then I noticed we don't show training for reviews in task views, I'm not sure _why_, perhaps for this exact reason. maybe we'll never know.

lennym: I feel like it batter to have the data and not need it, than to need it and not have it, so I'd probably dispense with this check and add the metadata to all PIL related tasks.

lennym: If there are some we _know_ we don't need (create?) then potentially filter those out.

PR #283, opened 2020-06-24
Improve performance of task uniqueness check

lennym:
Use a `whereIn` instead of a `whereNotIn` for finding open tasks. This offers vastly improved performance because it facilitates the use of the index on task status.

### DIFF
--- a/test/helpers/asl-db.js
+++ b/test/helpers/asl-db.js
@@ -6,6 +6,7 @@ module.exports = settings => {
     init: (populate, keepAlive) => {
       const schema = Schema(settings);
       const tables = [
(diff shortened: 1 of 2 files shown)


test/helpers/asl-db.js
```diff
@@ -6,6 +6,7 @@ module.exports = settings => {
     init: (populate, keepAlive) => {
       const schema = Schema(settings);
       const tables = [
+        'AsruEstablishment',
```
lennym, line 9: Not 100% sure what state I got in to require this, but my test run blew up on 100% of tests without it.

PR #284, opened 2020-06-26
Include tasks actioned by the user in the profile related task results

wheelsandcogs:


test/integration/related-tasks/profile-touched.js
```diff
@@ -0,0 +1,168 @@
...
+          'discarded by asru'
+        ];
+
+        return request(this.workflow)
+          .get(`/related-tasks?model=profile-touched&modelId=${asruSuper.id}&limit=10000`)
+          .expect(200)
+          .expect(response => {
+            assertTasks(expected, response.body.data);
+          });
+      });
```
wheelsandcogs, line 126: The test above is the only new test - the rest have just been re-organised cause that test file was mammoth.

test/data/activity.js
```diff
@@ -0,0 +1,25 @@
...
+};
+
+const activity = [
+  {
+    id: uuid(),
+    caseId: ids.task.project.discardedByAsru,
+    changedBy: asruSuper.id,
+    eventName: 'status:with-inspectorate:discarded-by-asru',
+    event: {},
+    ...generateDates(34)
```
wheelsandcogs, line 21: not strictly needed but should make the data seem a bit more realistic as it tallies with the dates on the task itself.

PR #285, opened 2020-06-29
Feature - submitted to asru decorator

joefitter:
Added decorator which checks if a task has been with-intspectorate or with-licensing


lib/api.js
```diff
@@ -65,6 +65,7 @@ module.exports = settings => {
   flow.decorate(decorators.filterComments(settings), { list: false });
   flow.decorate(decorators.deleteComments(settings), { list: false });
   flow.decorate(decorators.activityLog(settings), { list: false });
+  flow.decorate(decorators.submittedToAsru(settings), { list: false });
```
lennym, line 68: Does this work in the context that it needs with `list: false`?

PR #287, opened 2020-07-06
Add hook to re-index search when entity changes

lennym:
If a task relating to a profile, project or establishment completes then trigger a request to the search service to re-index the record.

Note: not waiting on any code changes, but probably doesn't need to ship until we have search working at least on preprod.

### DIFF
--- a/lib/middleware/profile.js
+++ b/lib/middleware/profile.js
@@ -2,7 +2,13 @@ module.exports = settings => {
   const { Profile } = settings.models;
   return (req, res, next) => {
 
-    delete req.user.access_token;
+    // define access token as a non-enumerable property so it is accessible in code downstream
+    // but is not enumerated in activity logs event data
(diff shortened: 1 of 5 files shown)


lib/middleware/profile.js
```diff
@@ -2,7 +2,13 @@ module.exports = settings => {
...
   return (req, res, next) => {
 
-    delete req.user.access_token;
+    // define access token as a non-enumerable property so it is accessible in code downstream
+    // but is not enumerated in activity logs event data
+    const token = req.user.access_token;
+    Object.defineProperty(req.user, 'access_token', {
+      enumerable: false,
+      value: token
+    });
```
joefitter, line 11: good plan

lennym: There's probably large chunks of the data that ends up in activity logs that should get the same treatment tbh.

PR #288, opened 2020-07-10
Refresh establishment search index when ASRU assignments change

lennym:
Otherwise the assignments shown in the search results will not be accurate until the next time the indexer runs.

### DIFF
--- a/lib/hooks/search/index.js
+++ b/lib/hooks/search/index.js
@@ -2,14 +2,18 @@ const { get } = require('lodash');
 
 module.exports = settings => event => {
   if (settings.search) {
-    const allowed = ['establishment', 'profile', 'project'];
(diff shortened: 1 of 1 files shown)


lib/hooks/search/index.js
```diff
@@ -2,17 +2,22 @@ const { get } = require('lodash');
...
+    let id = get(event, 'data.id');
+
+    if (model === 'asruEstablishment') {
+      model = 'establishment';
+      id = get(event, 'data.data.establishmentId');
+    }
 
     // note the extra `s` on the model name
     const url = `${settings.search}/${model}s/${id}`;
+    console.log(url);
```
joefitter, line 20: intentional?

lennym: No.

PR #291, opened 2020-07-17
Add hook for resolving draft ppl transfers

wheelsandcogs:


lib/hooks/create/project.js
```diff
@@ -135,6 +135,16 @@ module.exports = settings => {
...
           });
 
+      case 'transfer-draft':
+        return Promise.resolve()
+          .then(() => Project.query().findById(id))
+          .then(project => {
+            if (project.status !== 'inactive') {
+              throw new BadRequestError(`cannot perform 'transfer-draft' action for a non-draft project`);
+            }
+            return model.setStatus(resolved.id);
```
lennym, line 145: I think these are fine to autoresolve. I can't see a good reason to leave a task behind for a fairly straightforward edit to a draft PPL.

wheelsandcogs: Sorted.

PR #298, opened 2020-08-07
Bugfix - normalise training

joefitter:
Added decorator to nomalise certificates/exemptions of in flight tasks

### DIFF
--- a/lib/decorators/normalise-training.js
+++ b/lib/decorators/normalise-training.js
@@ -36,9 +36,9 @@ function flattenModules(certificate) {
   };
 }
 
-function normaliseTraining(c) {
-  const certificates = (c.certificates || []).map(flattenModules);
-  const exemptions = normaliseExemptions(c.exemptions || []);
+function normaliseTraining(data) {
+  const certificates = (data.certificates || []).map(flattenModules);
+  const exemptions = normaliseExemptions(data.exemptions || []);
 
   return [
     ...certificates,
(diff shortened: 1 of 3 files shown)


lib/decorators/normalise-training.js
```diff
@@ -0,0 +1,57 @@
...
+  const exemptions = normaliseExemptions(c.exemptions || []);
+
+  return [
+    ...certificates,
+    ...exemptions
+  ];
+}
+
+module.exports = c => {
+  if (c.certificates || c.exemptions) {
```
lennym, line 50: I'm pretty sure this is `c.data.certificates` etc.

lennym: Also ~c~ `task`

joefitter: right you are, I haven't actually tested this :)

lib/decorators/normalise-training.js
```diff
@@ -0,0 +1,60 @@
...
+  const exemptions = normaliseExemptions(data.exemptions || []);
+
+  return [
+    ...certificates,
+    ...exemptions
+  ];
+}
+
+module.exports = task => {
+  if (task.data.certificates || task.data.exemptions) {
```
lennym, line 50:
```suggestion
  if (task.data && (task.data.certificates || task.data.exemptions)) {
```

PR #303, opened 2020-08-14
Add certificates to project tasks on submission

lennym:
Training needs to be shown on PPL applications as it was at the point of submission rather than being displayed live.


joefitter:
it is, it's taken from the version data, not workflow

PR #311, opened 2020-09-16
Add days since deadline prop in decorator

wheelsandcogs:


lib/decorators/deadline/get-deadline.js
```diff
@@ -9,6 +9,8 @@ const EXTENDED_DEADLINE = 55;
 // configure bank holidays
 moment.updateLocale('en', { holidays: bankHolidays });
 
+const daysSinceDate = (date, from = new Date()) => moment(from).diff(moment(date), 'days');
```
lennym, line 12: If this is never called with a second argument then it's probably a lot easier to omit it completely
```suggestion
const daysSinceDate = (date) => moment().diff(moment(date), 'days');
```

lennym: But also isn't this intended to be based on the activity date, not the current date?

wheelsandcogs: The decorator does not run against the task data stored in `activity_log.event.data`, so this is only useful for the current deadline.

We'll still have to calculate the `daysSinceDeadline` on the UI side for activity log entries, but we'll use the deadline stored against the activity rather than the current deadline (fix in another PR).

PR #312, opened 2020-09-17
Feature - cat e pils

joefitter:
* Added create hook for trainingPil - creates trainingPil  and sends for endorsement
* Added endorsed hook - to est-b if existing pil elsewhere
* Added decorator to include trainingPil status
* Added training pil related tasks list

### DIFF
--- a/lib/api.js
+++ b/lib/api.js
@@ -74,10 +74,10 @@ module.exports = settings => {
   flow.hook('pre-create', hooks.unique(settings));
   flow.hook('create', hooks.updateData(settings));
   flow.hook('create', hooks.meta(settings));
+  flow.hook('create', hooks.modelData(settings));
   flow.hook('create', hooks.updateAction(settings));
   flow.hook('create', hooks.submissionMeta(settings));
   flow.hook('create', hooks.create(settings));
-  flow.hook('create', hooks.modelData(settings));
   flow.hook(`status:*:${resubmitted.id}`, hooks.updateAction(settings));
   flow.hook(`status:*:${resubmitted.id}`, hooks.submissionMeta(settings));
   flow.hook(`status:*:${resubmitted.id}`, hooks.create(settings));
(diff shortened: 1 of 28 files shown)


lib/api.js
```diff
@@ -77,6 +78,7 @@ module.exports = settings => {
   flow.hook('create', hooks.updateAction(settings));
   flow.hook('create', hooks.submissionMeta(settings));
   flow.hook('create', hooks.create(settings));
+  flow.hook('create', hooks.modelData(settings));
```
lennym, line 81: I don't think this should be here, since it's already applied on L77
```suggestion
```

lib/decorators/training-pil-meta.js
```diff
@@ -0,0 +1,33 @@
...
+    const openRevocation = !!parseInt(revocationCount.count, 10);
+
+    return {
+      ...c,
+      openRevocation,
+      modelStatus: trainingPil && trainingPil.status
+    };
+  };
+
+};
```
lennym, line 33: Can you rename `c` to `task` everywhere here please?

test/integration/task-lists/ntco.js
```diff
@@ -30,7 +30,8 @@ describe('NTCO', () => {
     it('sees tasks with status "with-ntco" or "awaiting-endorsement" for their own establishment', () => {
       const expected = [
         'pil with ntco',
-        'another with-ntco to test ordering'
+        'another with-ntco to test ordering',
+        'trainingPil with ntco'
```
lennym, line 34: It might be useful to have something to cover `isTrainingPilAtOtherEst` here as well.

lib/api.js
joefitter: ah - no this needs to happen after create, as the training pil won't exist until then. I left the other one before create as I didn't know what it would break elsewhere

lennym: Let's find out then and remove the other. Running the same hook twice is nasty.

PR #314, opened 2020-09-24
Add a script to patch in task deadlines where required

wheelsandcogs:


lennym:
Could you wrap it all in a transaction please?

bin/add-deadlines-to-legacy-tasks
```diff
@@ -0,0 +1,82 @@
...
+};
+
+const migrate = async () => {
+  const db = Database.connect(settings.taskflowDB);
+  Task.db(db);
+
+  // only need to patch tasks that have already been submitted, because the deadline object gets added on submission
+  const withAsruOrComplete = withASRU().concat(['resolved', 'rejected']);
+
+  const query = Task.query()
```
lennym, line 61:
```suggestion
  const tasks = await Task.query()
```

bin/add-deadlines-to-legacy-tasks
```diff
@@ -0,0 +1,82 @@
...
+    .select('*')
+    .whereIn('status', withAsruOrComplete)
+    .whereJsonSupersetOf('data', { model: 'project', action: 'grant' })
+    .whereRaw("data->>'deadline' is null"); // will match both deadline: null and missing deadline key
+
+  const tasks = await query;
```
lennym, line 67:
```suggestion
```

bin/add-deadlines-to-legacy-tasks
wheelsandcogs: This was because I was debugging the query sql 👍

bin/add-deadlines-to-legacy-tasks
```diff
@@ -0,0 +1,84 @@
...
+    .whereJsonSupersetOf('data', { model: 'project', action: 'grant' })
+    .whereRaw("data->>'deadline' is null"); // will match both deadline: null and missing deadline key
+
+  console.log(`found ${tasks.length} tasks with no deadline property set`);
+
+  for (const task of tasks) {
+    await patchTaskDeadline(task.id, transaction); // process the tasks one at a time to avoid ballooning memory
+  }
+
+  await transaction.commit();
```
lennym, line 75: The transaction will need to be rolled back in case of an error.

PR #316, opened 2020-09-24
Deadline passed checker

wheelsandcogs:
This should be set to run after midnight each night, to flag any tasks where the deadline has passed.

### DIFF
--- a/bin/deadline-has-passed
+++ b/bin/deadline-has-passed
@@ -5,10 +5,6 @@ const Database = require('@ukhomeoffice/taskflow/lib/db');
 const Task = require('@ukhomeoffice/taskflow/lib/models/task');
 const settings = require('../config');
 
-const setHasPassed = async (task, transaction) => {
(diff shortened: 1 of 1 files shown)


bin/deadline-has-passed
```diff
@@ -0,0 +1,59 @@
...
+          .whereRaw("data->>'deadlinePassed' is null") // hasn't yet been marked as passed
+          .as('cases');
+      })
+      .where('deadlineDate', '<', today);
+
+    console.log(`found ${tasks.length} tasks that have passed their deadlines`);
+
+    for (const task of tasks) {
+      await task.$query(transaction).patchAndFetch({ data: { ...task.data, deadlinePassed: true } });
+    }
```
lennym, line 43: It doesn't much matter here because there's never likely to be more than a very small number, but generally I'd tend towards doing this update in a single query rather than iterating over a set.

PR #318, opened 2020-09-28
Don't refresh updatedAt when patching tasks

wheelsandcogs:


bin/add-deadlines-to-legacy-tasks
```diff
@@ -36,20 +38,27 @@ const patchTaskDeadline = async (taskId, transaction) => {
...
     }
     return deadline;
   }, {});
 
   if (isEmpty(deadline)) {
-    console.log(`${task.id}: no deadline required`);
-    return Promise.resolve();
+    if (task.data.deadline) {
+      stats.removed++;
+      return task.$query(transaction).patchAndFetch({ data: { ...task.data, deadline: null } }).context({ preserveUpdatedAt: true });
```
wheelsandcogs, line 51: This bit is here to remove the partial deadlines (e.g. `{"isExtended":false,"isExtendable":true}`) that got added to the task records on preprod the first time we ran the script.

bin/add-deadlines-to-legacy-tasks
```diff
@@ -36,20 +38,27 @@ const patchTaskDeadline = async (taskId, transaction) => {
         return {}; // submission was not complete and correct, reset deadline
       }
     }
-    if (activity.eventName === 'update') { // possible deadline extend event
+    if (activity.eventName === 'update' && !isEmpty(deadline)) { // possible deadline extend event
```
wheelsandcogs, line 41: This is to prevent the partial deadlines getting added (e.g. `{"isExtended":false,"isExtendable":true}`).

bin/add-deadlines-to-legacy-tasks
```diff
@@ -36,20 +38,27 @@ const patchTaskDeadline = async (taskId, transaction) => {
...
   if (isEmpty(deadline)) {
-    console.log(`${task.id}: no deadline required`);
-    return Promise.resolve();
+    if (task.data.deadline) {
+      stats.removed++;
+      return task.$query(transaction).patchAndFetch({ data: { ...task.data, deadline: null } }).context({ preserveUpdatedAt: true });
+    } else {
+      stats.noop++;
+      console.log(`${task.id}: no deadline required`);
+      return Promise.resolve();
```
wheelsandcogs, line 55: Same as before, this is when there is no deadline to add.

bin/add-deadlines-to-legacy-tasks
```diff
@@ -36,20 +38,27 @@ const patchTaskDeadline = async (taskId, transaction) => {
...
+    }
   }
 
+  stats.added++;
   console.log(`${task.id}: ${JSON.stringify(deadline)}`);
-  return task.$query(transaction).patchAndFetch({ data: { ...task.data, deadline } });
+  return task.$query(transaction).patchAndFetch({ data: { ...task.data, deadline } }).context({ preserveUpdatedAt: true });
```
wheelsandcogs, line 61: don't touch the updated at column.

bin/add-deadlines-to-legacy-tasks
```diff
@@ -65,9 +74,13 @@ const migrate = async () => {
       .select('id')
       .whereIn('status', withAsruOrComplete)
       .whereJsonSupersetOf('data', { model: 'project', action: 'grant' })
-      .whereRaw("data->>'deadline' is null"); // will match both deadline: null and missing deadline key
+      .where(builder => {
+        // will match deadline: null, missing deadline key, and missing deadline->standard key
+        builder.whereRaw("data->>'deadline' is null")
+          .orWhereRaw("data->'deadline'->>'standard' is null");
+      });
```
wheelsandcogs, line 81: Because preprod now has the partial deadlines on some records, we need to fetch them in order to fix them.

seeds/data/cases.json
```diff
@@ -143,8 +143,8 @@
       "action": "grant",
       "subject": "5b7bad13-f34b-4959-bd08-c6067ae2fcdd",
       "deadline": {
-        "standard": "2020-07-24",
-        "extended": "2020-08-14",
+        "standard": "2020-07-27",
+        "extended": "2020-08-17",
```
wheelsandcogs, line 147: The dates on this seed (and related activity) did not make sense for the submission date of `2020-06-01`. This is the correct deadline.

bin/add-deadlines-to-legacy-tasks
lennym: Does this script backfill all tasks that need a deadline?

If so then I'd probably get rid of any code that handles preprod being wonky and just run a manual query to remove all deadlines from preprod.

wheelsandcogs: yeah it does, ok let's do that.

wheelsandcogs: done.

PR #335, opened 2020-11-10
use moment business time to task bank hols into account

joefitter:


lennym:
This is a no-op because a working month is identical to a month

PR #349, opened 2021-01-13
Feature - RA

joefitter:
* Added actions for patch fork and grant for RA


lennym:
It would be helpful to get some basic level unit testing on this.

PR #355, opened 2021-02-05
Feature - CJSM

joefitter:
* Resolve update-cjsm task if from asru


lennym:
Could you add a unit test please?

PR #357, opened 2021-02-09
Add route to filter all tasks by query params

wheelsandcogs:


lennym:
Could you at add least one unit test that hits the endpoint and verifies it returns something meaningful?

Not looking for full coverage (unless you feel like it!) but a non-zero amount of coverage would be good.

lennym:
👍 

Thanks, that's great!

PR #360, opened 2021-02-12
Filter for change of licence holder

wheelsandcogs:


lib/query-builders/filters/by-action.js
```diff
@@ -1,3 +1,9 @@
 module.exports = action => query => {
+  if (action === 'change-licence-holder') {
+    return query.andWhere(builder => {
+      builder.whereJsonSupersetOf('data', { action: 'update' });
+      builder.whereRaw(`data->'modelData'->>'licenceHolderId' <> data->'data'->>'licenceHolderId'`);
```
lennym, line 5:
```suggestion
      builder.whereRaw(`data->'modelData'->>'licenceHolderId' != data->'data'->>'licenceHolderId'`);
```

It's equivalent, but one I immediately understand and the other breaks my brain.

PR #364, opened 2021-03-04
Feature - ROPS

joefitter:
* Skip workflow for rop create and update, and proc create update and delete.
* Respond with id when new rop created
* Bumped schema with rop models

### DIFF
--- a/lib/middleware/skip-workflow.js
+++ b/lib/middleware/skip-workflow.js
@@ -17,6 +17,12 @@ module.exports = settings => {
       if (model === 'profile' && action === 'updateLastLogin') {
         return true;
       }
+      if (model === 'rop' && action === ['update', 'create']) {
+        return true;
+      }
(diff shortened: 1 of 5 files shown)


lib/middleware/skip-workflow.js
```diff
@@ -17,6 +17,12 @@ module.exports = settings => {
       if (model === 'profile' && action === 'updateLastLogin') {
         return true;
       }
+      if (model === 'rop' && ['update', 'create', 'submit'].includes(action)) {
+        return true;
+      }
+      if (model === 'procedure' && ['update', 'create', 'delete'].includes(action)) {
+        return true;
+      }
```
lennym, line 25: We probably want to make sure these still work at all if the action isn't one of those.

I'd suggest a small modification to the create hook to set all `rop` and `procedure` models to `autoresolved` as a fallback.

joefitter: not a bad shout

PR #370, opened 2021-04-06
Tell resolver to delete the RA when ra task is discarded

wheelsandcogs:
needs https://github.com/UKHomeOffice/asl-resolver/pull/215

### DIFF
--- a/lib/hooks/discard/index.js
+++ b/lib/hooks/discard/index.js
@@ -6,6 +6,7 @@ module.exports = settings => {
   return model => {
     const type = get(model, 'data.model');
     const action = get(model, 'data.action');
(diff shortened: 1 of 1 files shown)


lib/hooks/discard/index.js
```diff
@@ -6,6 +6,7 @@ module.exports = settings => {
   return model => {
     const type = get(model, 'data.model');
     const action = get(model, 'data.action');
+    const changedBy = get(model, 'meta.user.profile.id');
```
wheelsandcogs, line 9: added this because the recall RA hook includes it, but it doesn't appear to be used anywhere other than the logs?

PR #401, opened 2021-07-13
Feature - recovered status

joefitter:
* Added recovered status, only usable by asru admin, autoforwards to with-inspectorate
* Dont include recovered in next steps - (decorator)

### DIFF
--- a/lib/decorators/next-steps.js
+++ b/lib/decorators/next-steps.js
@@ -1,5 +1,5 @@
 const getNextStepsForUser = require('../flow/get-next-steps-for-user');
-const { resubmitted, recovered } = require('../flow/status');
+const { resubmitted } = require('../flow/status');
 
 module.exports = settings => {
 
@@ -11,9 +11,6 @@ module.exports = settings => {
 
     return getNextStepsForUser(c, user.profile)
       .then(nextSteps => {
-        // recovered should not appear in next steps radios
-        nextSteps = nextSteps.filter(s => s.id !== recovered.id);
-
         if (c.data.model === 'project' && c.data.action === 'grant') {
           nextSteps = nextSteps.filter(s => s.id !== resubmitted.id);
         }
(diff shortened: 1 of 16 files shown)


lib/flow/get-next-steps-for-user.js
```diff
@@ -77,6 +77,10 @@ module.exports = (task, profile) => {
         nextSteps.push(discardedByAsru);
       }
 
+      if (!taskIsOpen(task) && userIsAsruAdmin(profile)) {
```
wheelsandcogs, line 80: Does this mean we can recover a granted task as well? Do we specifically want to check for `rejected` here?

lib/flow/status.js
```diff
@@ -139,6 +143,12 @@ module.exports = {
     id: 'rejected',
     commentRequired: true
   },
+  recovered: {
+    id: 'recovered',
+    commentRequired: false,
```
wheelsandcogs, line 148: Should we enforce an explanation, or do we think it's not necessary?

lib/flow/get-next-steps-for-user.js
joefitter: you're right - it does in client but it also wants to be more specific here - I'll update

lib/flow/status.js
joefitter: it is a button style form with a js.confirm dialog, I think that's fine as it's only going to realistically be done by Martin

lib/decorators/next-steps.js
```diff
@@ -11,6 +11,9 @@ module.exports = settings => {
 
     return getNextStepsForUser(c, user.profile)
       .then(nextSteps => {
+        // recovered should not appear in next steps radios
+        nextSteps = nextSteps.filter(s => s.id !== recovered.id);
```
wheelsandcogs, line 15: We don't do this for `discardedByAsru` which is added to the available steps in the same way and displays a similar UI - do we need it?

joefitter: yes or no, I do not know how that works

PR #402, opened 2021-07-16
Bugfix - recovered comments

joefitter:
* if a task is recovered, ignore this when working out if a comment is new or not

### DIFF
--- a/lib/decorators/filter-comments.js
+++ b/lib/decorators/filter-comments.js
@@ -18,6 +18,7 @@ module.exports = settings => {
         return a.createdAt < b.createdAt ? -1 : 1;
       });
 
+    // if a correction to a status is found then remove that correction as well as the preceding and following status change.
     let recoveries = Object.keys(pickBy(statusChanges, a => a.eventName === 'status:rejected:recovered')).map(i => parseInt(i, 10));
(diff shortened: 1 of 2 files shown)


lib/decorators/filter-comments.js
```diff
@@ -11,13 +11,20 @@ module.exports = settings => {
...
 
-    const statusChanges = c.activityLog
+    let statusChanges = c.activityLog
       .filter(a => a.eventName.match(/^status:/))
       .sort((a, b) => {
         // sort into ascending order by time
         return a.createdAt < b.createdAt ? -1 : 1;
       });
 
+    let recoveries = Object.keys(pickBy(statusChanges, a => a.eventName === 'status:rejected:recovered')).map(i => parseInt(i, 10));
```
lennym, line 21:
```suggestion
      // if a correction to a status is found then remove that correction as well as the preceding and following status change.
      let recoveries = Object.keys(pickBy(statusChanges, a => a.eventName === 'status:rejected:recovered')).map(i => parseInt(i, 10));
```

lennym: It took a fair amount of reading this code to work out WTF was going on.

PR #405, opened 2021-09-02
Allow filtering project tasks by type (amendment / transfer / ra / continuation / hasDeadline / etc)

wheelsandcogs:


lib/query-builders/filters/by-ppl-type.js
```diff
@@ -0,0 +1,28 @@
...
+      });
+
+    case 'hasDeadline':
+      return query.andWhere(builder => {
+        builder.whereRaw("data \\? 'deadline'");
+      });
+
+    case 'ra':
+      return query.andWhere(filterByAction('grant-ra'));
+  }
```
lennym, line 27: We may want to take a look at the performance of these in prod and add indexes if necessary.

PR #410, opened 2021-10-05
Feature - awerb change licence holder

joefitter:
* Updated change licence holder to require endorsement


lib/decorators/normalise-awerb.js
```diff
@@ -0,0 +1,41 @@
+const { mapValues } = require('lodash');
+
+function destupidify(value) {
```
lennym, line 3: 🍭

lennym:
You've added a new decorator file, but it's not actually being used anywhere.

joefitter:
ah this looks like a rebase error - I added this file in a different commit, destupidify should not be a thing 😬

PR #412, opened 2021-10-21
prevent events with no next status breaking the search refresh

wheelsandcogs:


lib/hooks/task-search/index.js
```diff
@@ -1,9 +1,16 @@
 const fetch = require('node-fetch');
+const { get } = require('lodash');
 const { autoForwards } = require('../../flow');
 
 module.exports = settings => event => {
   if (settings.search) {
-    if (event.status === 'autoresolved' || autoForwards(event.meta.next)) {
+    if (['new', 'autoresolved'].includes(event.status)) {
```
lennym, line 7: This still needs to be `event.meta.next`, we just need to check that it exists first.

When there are multiple hooks and multiple status changes in a single request then `event.status` may not be the status into which the status change event is transferring.

Example:
```
app.hook('status:foo:bar', e => e.setStatus('baz'));
app.hook('status:foo:bar', e => {
  console.log(e.status); // baz
  console.log(e.meta.next); // bar
});
```

PR #416, opened 2021-11-10
if returned PIL was with inspectorate return to inspectorate

wheelsandcogs:
If a PIL task is referred to an inspector, and then returned, then when it is endorsed it gets passed directly to the inspectorate by setting the status to `with-inspectorate`.

If this same task is then returned / recalled a second time, the last ASRU status will be `with-inspectorate` so we need to check for this as well.


lennym:
Can you add a test equivalent to [this one](https://github.com/UKHomeOffice/asl-workflow/blob/master/test/unit/hooks/endorsed/pil.js#L42-L55)?

lennym:
Also... good catch!

PR #421, opened 2021-12-15
WIP Add internal deadlines to PPL tasks on submission to ASRU

wheelsandcogs:


test/unit/hooks/set-internal-deadline/index.js
```diff
@@ -0,0 +1,217 @@
...
+        patch: sinon.stub()
+      };
+
+      const history = History(pplAmendmentResub);
+      history.status(withInspectorate.id);
+
+      const expected = {
+        internalDeadline: {
+          standard: moment().addWorkingTime(40, 'days').format('YYYY-MM-DD'),
+          extended: moment().addWorkingTime(55, 'days').format('YYYY-MM-DD')
```
wheelsandcogs, line 167: We can use the `task.data.deadline.isExtended` flag to decide which of these to use when rendering the deadline, it didn't seem sensible to track/manage the extension in multiple places.

PR #425, opened 2022-02-03
Add assignment to related tasks query

lennym:
This will include all tasks currently assigned to a particular ASRU user in the "Related tasks" section of their profile.

Also prevent a task from showing as related if assigning it to another user.

### DIFF
--- a/test/integration/related-tasks/profile-touched.js
+++ b/test/integration/related-tasks/profile-touched.js
@@ -116,9 +116,8 @@ describe('Related tasks', () => {
           });
       });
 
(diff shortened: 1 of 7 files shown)


test/integration/related-tasks/profile-touched.js
```diff
@@ -116,9 +116,8 @@ describe('Related tasks', () => {
           });
       });
 
-      it('includes tasks that have been actioned by the target profile', () => {
+      it('includes tasks created the the target profile', () => {
```
wheelsandcogs, line 119:
```suggestion
      it('includes tasks created by the target profile', () => {
```

PR #428, opened 2022-02-11
Add internal deadline hook and integration test

lennym:
Adding the integration tests for deadlines meant setting the project statuses in the seeds to appropriate values, which had some interesting side-effects - in particular a test that was effectively noop-ing for changing licence holders on stub licences because all the projects were `inactive`.

### DIFF
--- a/test/integration/hooks/deadlines.js
+++ b/test/integration/hooks/deadlines.js
@@ -1,0 +1,124 @@
+const request = require('supertest');
+const assert = require('assert');
+const moment = require('moment-business-time');
+const workflowHelper = require('../../helpers/workflow');
+const { holc } = require('../../data/profiles');
+const ids = require('../../data/ids');
+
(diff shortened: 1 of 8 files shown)


test/integration/hooks/deadlines.js
```diff
@@ -0,0 +1,126 @@
...
+    return workflowHelper.create()
+      .then(workflow => {
+        this.workflow = workflow;
+      });
+  });
+
+  beforeEach(() => {
+    return Promise.resolve()
+      .then(() => this.workflow.resetDBs())
+      .then(() => this.workflow.clearTaskList());
```
wheelsandcogs, line 19: if `resetDBs()` calls `tfdb.reset()`, why the second call to `clearTaskList()`?

lennym: Removed.

PR #454, opened 2022-08-23
ASL-4075 - Add nightly-jobs script and include project-refusal-deadline-passed

gaz1634:


wheelsandcogs:
Does the logging still appear as before?

wheelsandcogs (approved):
Approving on the assumption that the logging hasn't changed and still appears in the console.

gaz1634:
> 

It should at least log everything it did before as the exec callback is logging `stdout` from the called script to `console.log` and `stderr` to `console.error`. 
The only slight difference is there may be an additional log to say the job <job-name> failed, with some details of the failure. That way if the underlying script throws an error the details should also be logged.

wheelsandcogs:
👍

PR #455, opened 2022-08-24
Feature: Suspend / reinstate licences

wheelsandcogs:


lib/hooks/create/unique.js
```diff
@@ -23,7 +23,9 @@ module.exports = settings => {
       'establishment.update-conditions',
       'establishment.update-billing',
       'profile.resend-email',
-      'profile.confirm-email'
+      'profile.confirm-email',
+      'pil|project|establishment.suspend',
+      'pil|project|establishment.reinstate'
```
lennym, line 28: Are you _sure_ this won't match all pil and project tasks? I'd be tempted to replace with `*.suspend` and `*.reinstate` for a blend of simplicity and a certain answer to that question?

test/helpers/asl-db.js
```diff
@@ -1,41 +1,17 @@
...
-          }
-          return schema[table].query().truncate();
-        });
-      }, Promise.resolve())
+
+      const tableNames = Object.keys(schema)
+        .filter(key => !!schema[key].tableName)
+        .map(key => snakeCase(schema[key].tableName));
+
+      return schema.knex.raw(`TRUNCATE TABLE ${tableNames.join(', ')} CASCADE;`)
```
wheelsandcogs, line 14: Running a `TRUNCATE TABLE x CASCADE;` query per table in series appears to be considerably slower than just deleting the rows.

However, what if we just have a single SQL statement that truncates all the tables at once? 

If this is also slow then there might be the option to temporarily disable the FK constraints with 

```
ALTER TABLE x DISABLE TRIGGER ALL;
DELETE FROM x;
ALTER TABLE x ENABLE TRIGGER ALL;
```

lib/hooks/create/unique.js
lennym: Having tried this out a little using minimatch in CLI it seems to not match at all.

```
> match('pil.suspend', 'pil|project|establishment.suspend')
false
> match('pil.suspend', '+(pil|project|establishment).suspend')
true
```

It looks like the syntax needs to be `+(a|b|c).action`.

PR #463, opened 2022-10-27
Handle string dates

wheelsandcogs:


bin/profile-merge-log
```diff
@@ -42,7 +42,7 @@ const logProfileMerges = async () => {
         continue;
       }
 
-      const mergeTime = task.updatedAt.toISOString();
+      const mergeTime = typeof task.updatedAt === 'object' ? task.updatedAt.toISOString() : task.updatedAt;
```
lennym, line 45: `typeof null === 'object'`, so depending on the root cause this may not fix your problem.

```suggestion
      const mergeTime = task.updatedAt instanceof Date ? task.updatedAt.toISOString() : task.updatedAt;
```

PR #491, opened 2024-09-23
Asl 4512/CVE review

farrukhishaq:
upon merging last week, Drone split new vulnerabilities. Also, there is an issue with the script in the pipeline as the 

`npm run test:audit` has different results locally compared to the pipeline.


mohomeoffice (commented):
Farrukh can we hold on to these deployment till we pushed all the current changes to prep pod

mohomeoffice (approved):
Seems okay to me

PR #500, opened 2025-01-15
Upgrade/asl schema11.0.0

farrukhishaq:


mohomeoffice (commented):
process.chdir('../../node_modules/@asl/schema'); asl schema placed inside node_modules ?

farrukhishaq:
> process.chdir('../../node_modules/@asl/schema'); asl schema placed inside node_modules ?

fixed it, the ../.. was when i was using aspel workspace.

